PR #10, opened 2018-05-02
Fixed the Mangy Gnoll

Vasqu3z:
Will give this a once-over tomorrow when I have some more time. Hopefully should be the last outstanding fix but can't make any promises :/


TheGiddyLimit:
That's a 2,000 line fix. Looks like you've accidentally re-formatted a bunch of JSON?

PR #23, opened 2018-05-17
Adding Monster of the Guild

teen120:
First time doing this. Testing 2 monsters first. More to come later.


teen120:
I went ahead and did all that. Could you check and add it to the repo before this Sunday? I would like to use these in my session.

TheGiddyLimit:
Absolutely; I mostly approve things the same day they're posted. Timezone differences + working on MTF perhaps making it slower than it could have been.

PR #54, opened 2018-09-27
Create Modnar; Character Points.json

DMsGuild201:
Creation.
Variant rules for more engangement and alterante attunement system.


variantrule/Modnar; Character Points.json
```diff
@@ -0,0 +1,489 @@
+{
+	"_meta": {
+		"sources": [
+			{
+				"json": "Giddy",
```
TheGiddyLimit: This should be changed to `Modnar`, or something else unique to this source. 😁

variantrule/Modnar; Character Points.json
```diff
@@ -0,0 +1,489 @@
...
+				"abbreviation": "Brew",
```
TheGiddyLimit: Similarly, you might want to make this something mode identifiable. Not necessary, though.

variantrule/Modnar; Character Points.json
```diff
@@ -0,0 +1,489 @@
...
+				"version": "1.0.0",
+				"url": "https://www.gmbinder.com/share/-L8YS8M9xAdPhqY7d6tj",
+				"targetSchema": "1.0.0"
+			}
+		]
+	},
+
+	"type": "section",
+	"name": "Character Points [CP]",
+	"entries": [
```
TheGiddyLimit: Not quite sure what you're aiming for here -- if this is a single variant rule, it should be stored in an array with key `variantrule`. See other files for examples. Alternatively, I can easily fix it up, if you wish (and tell me which things are supposed to be individual rules).

PR #68, opened 2018-12-04
[Submission] Jonoman3000; Dark Arts Players Companion

tylermac:


TheGiddyLimit:
Very nice. This might be something to include as a "collection," rather than a "book," otherwise it'll only be visible on the books page* (adding it from the books page will cause all the other content to populate, however). Up to you, really. If the other content only makes sense in the context of the book, it should remain as-is.

* And the general homebrew management page.

tylermac:
Makes good sense to me. :)

PR #81, opened 2018-12-28
A few more items here

tylermac:
- Moved Compendium of Forgotten Secrets to Collection, and started adding in the other options from it.
- Fixed the fluff on the Griffin's Saddlebag and added the new posted items
- Added The Noble Class
- Added The Scholar Class


TheGiddyLimit:
Lovely, cheers

PR #83, opened 2019-01-02
Limitless Adventures Update

FrankD67:
Have added two encounters to Limitless Adventures.
Hopefully this works.

### DIFF
--- a/adventure/Limitless Adventures.json
+++ b/adventure/Limitless Adventures.json
@@ -3,7 +3,7 @@
 		"sources": [
 			{
 				"json": "Limitless Adventures",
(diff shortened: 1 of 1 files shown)


TheGiddyLimit:
Looks good-ish, although something's up with the entire file being marked as a diff. Have you changed the encoding/line feeds?

Might be worth doing a
`git config --global core.autocrlf true`
for future.

FrankD67:
Fucked up my first pull request oh Well

I updated the whole file, also noticed on "abbreviation": "Las", so changed it to "abbreviation": "LAS",

PR #88, opened 2019-01-16
Deep Magic 13: Dragon Magic

eberron94:
Added Deep Magic 13: Dragon Magic spells
subclass and feats not added


eberron94:
Added to index.json wrong, will open new pull request with fixed commit

PR #89, opened 2019-01-16
Added Deep Magic 13: Dragon Magic to Collection

eberron94:
only added the spell section of the book


TheGiddyLimit:
Don't worry about the collections index, it's generated by a script. 

For future reference, `"dateAdded"` should be in seconds, not milliseconds.

PR #90, opened 2019-01-16
Various updates (marked individually)

DMsGuild201:
You can ignore - 
class/Robert J Schwalb; Call to Arms The Warlord.json

that's my mistake. but is forthcoming from L0neGamer


TheGiddyLimit:
Changes applied. Tyvm.

PR #103, opened 2019-01-31
NPC updated 

ghost:
Used new format for Fluff, only about 20% done in the reformating.


DMsGuild201:
additional Typos and another file to push

PR #108, opened 2019-02-02
Created Zarieth; The Witch

Piefanity:
Adds the Witch Class by reddit.com/u/Zarieth
includes the following

- 1 Witch Class
- 5 subclasses for the Witch
- Witch's Spellist
- Unique SpiritBindings as Class Feature


class/Zarieth; The Witch
```diff
@@ -0,0 +1,1836 @@
...
+[
+	{
+		"type": "optfeature",
+		"name": "Agony",
+		"source": "The Witch",
+		"page": 10,
+		"featureType": "Spirit Binding",
+		"entries": [
+			"In the presence of a spirit of Agony, foes’ wounds continue to bleed and fires continue to burn. While Agony is bound, for every damage die you roll that shows a 1, you may treat the roll as the maximum possible number for that die for a number of dice up to the level of the spirit.",
+			"{@b {@i Release.}} You may release Agony as part of any attack or spell which deals damage. In addition to the damage, you may roll an extra number of equivalent dice equal to the level of the spirit. If the attack or spell uses more than one type of die, choose the largest. This extra damage is psychic damage. Additionally, all 1s are treated as the maximum value for each die."
```
TheGiddyLimit: This is better represented as:
```json
{
  "type": "entries",
  "name": "Release",
  "entries": [
    "You may release Agony as part of any attack or spell which deals damage. In addition to the damage, you may roll an extra number of equivalent dice equal to the level of the spirit. If the attack or spell uses more than one type of die, choose the largest. This extra damage is psychic damage. Additionally, all 1s are treated as the maximum value for each die."
  ]
}
```
(and the same for the others)

class/Zarieth; The Witch
```diff
@@ -0,0 +1,1836 @@
...
+	},
+	"class": 
+	[
+		{
+			"name": "The Witch",
+			"source": "The Witch",
+			"fluff": 
+			[	
+				{
+					"name": "",
```
TheGiddyLimit: You can (and should!) remove blank `name` strings

class/Zarieth; The Witch
```diff
@@ -0,0 +1,1836 @@
...
+		]
+	}
+
+],
+"optionalfeature": 
+[
+	{
+		"type": "optfeature",
+		"name": "Agony",
+		"source": "The Witch",
```
TheGiddyLimit: This should match your `"json"` source defined in `"_meta"`, e.g. be `"ZariethTheWitchFromGMBinder"` (and the same for the others)

class/Zarieth; The Witch
```diff
@@ -0,0 +1,1836 @@
...
+				"targetSchema": "1.0.0"
+			}
+		],
+		"dateAdded": 1549127108
+	},
+	"class": 
+	[
+		{
+			"name": "The Witch",
+			"source": "The Witch",
```
TheGiddyLimit: This should match your `"json"` source defined in `"_meta"`, e.g. be `"ZariethTheWitchFromGMBinder"` (and the same for the subclass sources)

class/Zarieth; The Witch
Piefanity: Changed for all the occurences. Wasn't sure how i could properly get the stylemark like this. Now I'll know for future conversions.

class/Zarieth; The Witch
Piefanity: Done. An Artifact of figuring out why the first line of my fluff wasn't properly showing. Removed it.

class/Zarieth; The Witch
Piefanity: Properly renamed them in all instances.

class/Zarieth; The Witch
Piefanity: Changed in all instances. Thanks for the Feedback.

Piefanity:
Resolved the Issues and just found out I forgot to add the proper file extension `.json` to it. Is there any quick fix without losing the change history?

PR #131, opened 2019-04-13
Fix sonaplayer; The Customizable Artificer

Azzurite:
I accidentally put `45` as cantrips available for level 15.

And a 5th/13th level feature that was there previously has been moved to lvl 9 / 15, so it should be removed at 5th / 13th level.


Azzurite:
Wait a sec, found one more thing...

PR #140, opened 2019-05-07
SRD, Homebrew Conversion Updates

DMsGuild201:


TheGiddyLimit:
FYI, files should be `.json` instead of `.JSON`, because Linux' filesystem is case-sensitive. If you're on Windows, the only way I know to get Git to catch on, is to rename the file to something random -> commit it -> rename it back to the correct, lowercase-extension, name -> commit it again.

DMsGuild201:
I am on windows. will do

TheGiddyLimit:
:+1: 
And of course, there are some merge conflicts to fix :joy:

TheGiddyLimit:
Lovely. Still a conflict in the SRD file, however?

DMsGuild201:
Yeah this one is the one we want... we know the one you have has errors... sorry.

Nekka pointed out that its just bad.

PR #180, opened 2019-06-29
Create SailorCat ; Graftling.json

ActualSailorCat:
I hope I did this correctly. I apologize if not. Adding a new homebrew race.


ActualSailorCat:
> 
> 
> Just FYI, the "cell-exact" stuff in rollable table columns gets automagically added if you just have numbers or number ranges in that column; there's plenty examples of this in the site's data. Either is technically correct, but your version is significantly more work to write 😉

Thanks a lot! I'll act on your advice going forward.

PR #190, opened 2019-08-01
Druid, fighter, monk, paladin, completed

Aarowaim:
Shortnames and sources moved for better legibility, many subclasses added.


Aarowaim:
Anything I can do for the next push to resurrect some whitespace?

TheGiddyLimit:
Use tabs instead of spaces, ty

Aarowaim:
Sure, I'll do so. Is it a concern with file sizes?

This is a bit of a tangent, but here's why my editor commits whitespace murder as the default setting.

Many platforms tabulate inconsistently. Among the different behaviours I've seen software give tabs:

* Aligning text to the next column with a multiple of 4
* Aligning text to the next column with a multiple of 8
* Indenting text by `x` columns, breaking any text that uses tab stops for formatting.
* Crashing commandline software or terminals that don't render or read control characters properly
* Renders no character

In comparison, spaces have a meaning that is consistent across the majority of platforms. Which synergizes well with open source endeavours where contributors may use a variety of different IDEs with different capabilities.

TheGiddyLimit:
File size is part of it, but worse is that it screws up the history, making it hard to see who edited what. You've essentially blanket-claimed the "last edit" of every line in the file.

This is hardly the place for a tabs vs. whitespace debate (which is as pointless as it is ancient, either way). We're using tabs here, and that's that. If you're using an IDE, you can chuck this `.editorconfig` in the root of the repo:
```
[*]
charset=utf-8
end_of_line=lf
trim_trailing_whitespace=true
insert_final_newline=true
indent_style=tab
indent_size=4
```
And that should get detected and used as the project default without you having to mess with any settings.

PR #193, opened 2019-08-09
Update aeyana; Mashiros Manual of Monstrous Magic.json

silverbass:
Also added the Inset and added SGT tags

Diff is probably big due to the tab/space thing


TheGiddyLimit:
Merge pls!

PR #217, opened 2019-09-05
Update class sonaplayer; The Customizable Artificer

Azzurite:
Simple update with some fixes of the author


Azzurite:
let me use the correct branch...

PR #254, opened 2019-11-18
added: entry values for armor for betterR20

Matsozetex:
acSpecial key doesnt put any text for betterR20 5eTools, entries are added to alleviate that. If you can add text of acSpecial for it that would be cool.


Matsozetex:
Fixed some values that should display correctly for better20 and 5etools, also clarified some items.

PR #255, opened 2019-11-22
Add Herbalism&Alchemy, but for real this time

Git-GoR:
_Feel free to remove the one you uploaded (without the V 1.2)_

Edit: Nvm, discovered that I can do that myself


TheGiddyLimit:
🤔 I specifically removed the "v1.2" part, though. Don't really like having it in the source "full" text either, since it's supposed to be kept in the `version` field, but I let that slide. Is there any reason for it to be in the source/file name?

Git-GoR:
No no, it's just that as you added the file yourself, I didn't know if github would recognize it as the same file from my fork and "update" it, or as a different file and keep both of them. That's why I erased the old one, just to be sure. 

The name doesn't matter, keep it consistant with the others.

PR #261, opened 2019-11-29
Add files via upload

Dezolis:


TheGiddyLimit:
Sorry, what is this supposed to be? You're PRing a HTML file as .json? If I open it as a HTML file, it's an exact copy of what's already in the repo, under `collection`?

PR #264, opened 2019-12-05
Ranges fixed

Matsozetex:
Added "A" properties to ranged weapons to display ranges.
Added "T" properties to thrown weapons to display ranges.
Previously had just used custom properties for these, but it seems the schema requires the in-built properties to be used to display ranges. It would be cool if there was a work around.

### DIFF
--- a/item/TheApoApostolov; Items for Grit and Glory.json
+++ b/item/TheApoApostolov; Items for Grit and Glory.json
@@ -2454,7 +2454,8 @@
 			"source": "GAG",
 			"property": [
 				"LITE",
-				"LIMT"
+				"LIMT",
+				"THRO"
 			],
 			"entries": [
 				"Simple Thrown Weapons"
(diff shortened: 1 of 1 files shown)


TheGiddyLimit:
Ah yes, I intended to fix that alongside having `dmg2` render without needing to specify things as "versatile". Will do that now. You can probably roll this back (as well as the versatile stuff? Up to you).

Matsozetex:
I will get to it on the weekend, but for now take it as a intermediate fix until it is updated.

TheGiddyLimit:
Sure, sure. It's not actually that simple either way; I think I'll need to add some template system to allow you to include whatever data you want when it renders. As a potential example:
```json
// Custom item property
...
"name": "Custom Thrown"
"template": "{{current_render}} thrown ({{item.range}})"
...
// Custom item property
...
"name": "Alt Damage"
"template": "{{current_render}} ({@dice {{item.dmg2}}})"
```
Thoughts? Things without specific templates would render as they do currently; just displaying the name.

Matsozetex:
Looks good, I mean, anything that I could semi-properly do is pretty good.

PR #275, opened 2020-01-06
Add new entries to item, spell, deity, and creature

hafdon:


hafdon:
I used the same `_meta` prop values for all four `.json` files, and each of the files has the same name (`'hafdon; zorq.json'`). Let me know if these need to be unique so as not to generate conflicts.

TheGiddyLimit:
Do you want them to be individual files? You could merge them all together into a `collection` instead. If not, I don't mind; identical sources shouldn't be a problem.

hafdon:
Ooh, I didn't realize I could do a collection. Yes, I'll add a commit to merge them into a collection and create a pull for that.

PR #311, opened 2020-02-05
Creators Request

MrVauxs:


TheGiddyLimit:
Not just ignoring this, for what it's worth (and/or if you're in touch with the creator). Going to add an "unlisted" feature to remove them from the site rather than delete the files, as (A) would rather not throw away perfectly good conversion work, and (B) they'll still be in the Git history anyway.

TheGiddyLimit:
File has been unlisted.

PR #323, opened 2020-02-23
resolve duplication & add Class/race

SnortsAlot:
remove some duplication in Abrev IDs
add Lingering Soul


.gitignore
```diff
@@ -2,3 +2,5 @@
 .vscode/
 .vs/
 node_modules/
+*.jpg
+*.png
```
TheGiddyLimit: Do not add image exclusions

SnortsAlot: fixed

PR #324, opened 2020-02-24
update to giddy; hook of butchery

SnortsAlot:
included itemproperties aspect.


TheGiddyLimit:
The diff is now only whitespace, is this the intent?

PR #326, opened 2020-02-24
ADD: New subraces to Flore; Explorer's guide to Xhoria

Moonlington:
Also realized my "dateAdded" was incorrect, so i changed it to the correct time it was added


Moonlington:
Have to change a few things before merging

Moonlington:
Made the changes that I wanted to make

PR #335, opened 2020-03-06
Fix Blood Hunter 2020 JSON

Smashman:
Subclasses didn't have source causing class page to fail to load
Blood Hunter class is separate to Critical Role so source abbreviation was incorrect


TheGiddyLimit:
This looks identical to the other PR I just merged, so... closing?

PR #336, opened 2020-03-06
Updated Blood Hunter

MrVauxs:
Different "json" and "abbreviation" to BH2020 instead of CR2020
Sources on subclasses
A bit of formatting


Smashman:
You're welcome.

PR #398, opened 2020-04-16
Fix broken links, update image directory urls

Spapz:
Just under 1000 errors from my crusade. :(
Also changed all SaF image URLs to the one without 3pp.
Added a few footnotes to links to Kingdoms & Warfare, which is unconverted.

@TheGiddyLimit: please rename `_img/SaF (3pp)/` to `_img/SaF/`.


Spapz:
Corrected the rest of the 3pp image folder references, but still can't change the directories myself because Github is evil.

Assorted other minor fixes otherwise.

PR #426, opened 2020-05-08
Spappz edits

Spapz:
Uploading CftD images because there's gonna be so many of them and I don't want to deal with jargon URLs like /ajF34n2.png

Also converted CftD and Underworld images to webp because it's objectively better. I'll hold off the crusade for now. 😎


TheGiddyLimit:
That's a negatory on the webp; until we can open them in MS Paint, it's not a valid image format

Spapz:
Weird criteria, but it works for me. 🤔 

https://i.vgy.me/3YMmQD.mp4

TheGiddyLimit:
I don't really care. We're not using it either way.

PR #435, opened 2020-05-11
Fix font URLs

Spapz:
I screwed up guessing things. 😔

### DIFF
--- a/collection/Hit Point Press; Humblewood Campaign Setting.json
+++ b/collection/Hit Point Press; Humblewood Campaign Setting.json
@@ -42,10 +42,10 @@
 				}
 			],
 			"fonts": [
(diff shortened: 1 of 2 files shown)


TheGiddyLimit:
Oh, that's a commit hash huh. Replace them all with `master` (e.g. `3c1d24a05944f93542c1b0a1955efbdd4f92e5a9` -> `master`) and if that works, you can just use it always instead

Spapz:
Strange, given that Humblewood's existing and functional fonts (which is what I'm copying) has those hashes. That said, using `master` does make functional links, so I guess you're correct. I'll change all the relevant links in both brews.

TheGiddyLimit:
Well it has the hash of the commit in which you added it, and you added the new fonts in a different commit, no?

Spapz:
Yeah, I mean that Ragu, who did the original Humblewood conversion, had a whole bunch of font links _with_ commit hashes instead of `master`, thus my confusion why I couldn't do that either. Guess it's a matter of committing it to you and _then_ updating the link to match. ¯\_(ツ)_/¯

PR #462, opened 2020-06-06
Parser Error resolution

SnortsAlot:
Fix files for homebrew support for 3rd party support.


TheGiddyLimit:
`Modnar; Modnars Marvelous Miscellany - Tables & Trinkets.json` is invalid JSON
`collection/Baker; MorgraveMiscellany.json` uses "MoMi" instead of the full JSON source as `"source"` in some entities

PR #463, opened 2020-06-06
More Parser Compatibility

SnortsAlot:
Fix files for homebrew support for 3rd party support.

AND 
More files
AND
Fixes for broken JSON 
AND
Removing incomplete files


SnortsAlot:
Ok, hopefully that is fixed

PR #495, opened 2020-06-20
MrVauxs finally getting Git on Desktop instead of nuking his forks every day at midnight

MrVauxs:


TheGiddyLimit:
What's with all the page removals?

MrVauxs:
Quote from Discord:
Your script auto-assumed that all subclasses inherit the same page as the class itself
Which was false, considering Bloodrager for example does not exist on the page 46 of... anything

PR #519, opened 2020-07-05
Fix mfov Slime Domain

Git-GoR:
Regarding the wrong `{@d` tags for dice and `{@c` for cantrips, were they a thing a while ago?
Because I fixed similar ones before, and someone mentioned it in the Discord. I've never used them myself, so I wouldn't know.


TheGiddyLimit:
no lol

PR #543, opened 2020-07-15
Devikyn; Variant Sorcerer

MrVauxs:
Class: Done
Subclasses: 6/6 Done
Optional Features: 12/12 Done
Typos: (Hopefully) Squashed
Tags: Embedded

### DIFF
--- a/class/devikyn; Variant Sorcerer.json
+++ b/class/devikyn; Variant Sorcerer.json
@@ -1,0 +1,1947 @@
+{
+	"$schema": "../schema.json",
+	"_meta": {
+		"sources": [
+			{
+				"json": "VariantSorcerer_By_devikyn",
+				"abbreviation": "VS",
+				"full": "Variant Sorcerer",
+				"authors": [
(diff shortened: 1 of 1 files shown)


class/devikyn; Variant Sorcerer.json
```diff
@@ -0,0 +1,3551 @@
...
+				]
+			}, 
+			"classTableGroups": [
+				{
+					"colLabels": [
+						"Metamagics Known"
+					], 
+					"rows": [
+						[
+							"—"
```
TheGiddyLimit, line 91: Why not use `0`? Are you using a specific dash?

MrVauxs: That's how it is shown in the pdf, and hey, Monks Ki Points, Normal Sorcerers Sorcery Points, etc. also have such formatting.

TheGiddyLimit: Yes, but `0` gets rendered as a dash too, and is better for automation

PR #589, opened 2020-08-21
Update NEThingMagic; Card Slinger.json

TooTallTye:


TooTallTye:
made more changes and final update as of 0200 est on 8/25/2020

PR #613, opened 2020-09-16
Add DDEX1-04: Dues for the Dead

surged20:


TheGiddyLimit (approved):
Generally, content from one source should be kept in a single "collection" file. Unless you've got a good reason to keep the adventure/creatures split, I'll merge these together.

PR #614, opened 2020-09-17
chore: clean/index/update

BucketListMod:


TheGiddyLimit (approved):
I would prefer it if you did not run the clean/index scripts in a PR, as these get batch-run over however many submissions appear in a day or two. Each commit author running them in each PR simply generates merge conflicts/even more file churn than the scripts naturally generate.

BucketListMod:
Oh, I presumed that these were the responsibility of the user who submitted
the file in the first place, as I got pinged for typos in Exploring Eberron
multiple times, and I didn't know any other way to fix them. I'll refrain
in the future.

>

PR #714, opened 2020-12-05
ToB1+2 Typos & Monster Images

Spappz:


TheGiddyLimit (approved):
I see you have chosen to break every VTT user's token images by changing the paths

Spappz:
Completely evaded my mind; the subsequent complaints on the server are 100% my bad. Hopefully the new full-size images will be a small pittance in exchange for the inconvenience 😔

TheGiddyLimit:
It is what it is. I could've equally had you move all the info images into their own directory, and have the token images be in the old place

PR #740, opened 2021-01-08
Upload of the modified Talent Tree

Maalunar:
Updated for Tasha feats and with extra tools talents. Leaving the original brew untouched.


Maalunar:
Renaming it to "Talent Trees (modified)" instead of the modified original "Talents Tree and Tools", for clarity.

TheGiddyLimit (approved):

PR #752, opened 2021-01-25
Recipes 

SnortsAlot:
Add collection/modnar's cooking and brewing - contains homebrew recipes now.

collection of drinks


TheGiddyLimit:
Seems you've got a duplicate file there

PR #815, opened 2021-03-01
Finished Arcadia Issue 1 conversion

ineedanamepls:
- Finished up Arcadia Issue 1 conversion and reuploaded it
- Added images for current/future conversions


TheGiddyLimit (changes requested):
You've stuck an extra PDF + JSON file in your image directory, neither of which should be there.

ineedanamepls:
sorry about that, fixed i think :)

PR #878, opened 2021-03-24
JSON released, also comes with art and token.

LadySaga:
Removed and updated a bunch of typoed art filenames, added missing tokens, and uploaded the actual json file for the creature pack.


TheGiddyLimit (approved):

TheGiddyLimit:
For future reference, filenames should be `<author>; <name>.json`, i.e. `michifromkmk; The Fluffy Folio.json` here

Issue #953, opened 2021-05-19
Request homebrew upload


Fuffy-Katons:
I understand that if I would like my homebrew to be uploaded to the 5e.tools website I should make an issue here so that is what I'm doing. Can I please get this homebrew uploaded. below is a PasteBin link and raw. 

https://pastebin.com/raw/Gn0P4GmC

https://pastebin.com/Gn0P4GmC

TheGiddyLimit:
Almost--it's preferred if you make a Pull Request to directly add the files. If you click "Add File" at the top of the repo, it should guide you through.

Fuffy-Katons:
ok ill do that thx

PR #957, opened 2021-05-20
Bogdan Max; Bogo's Homebrews

Fuffy-Katons:
I think this is what should be done if I want my homebrew to be uploaded to the 5e.tool site. I see that the ones already in the repo are categorized by what they have most in them so if this is useful mine has mostly items.


TheGiddyLimit:
You forgot the `.json` file extension, but, good enough.

Closes #953 .

PR #969, opened 2021-05-25
Add Sigilant Class by rsquared

DefendedPlains:
Created the Sigilant class (a rune based mage knight class) by rsquared including its four subclasses and its homebrew spells.


DefendedPlains:
Closing to make updates then resubmit.

PR #986, opened 2021-06-05
Add Ghostfire Gaming's Grim Hollow Subclasses

rannieryjesuino:


Lxran:
The whole book's conversion is also complete and in the PR list, this is not needed.

rannieryjesuino:
@Lxran thanks alot!

PR #994, opened 2021-06-09
Add InxSinon; Tera's Travels At the Peaks of Gravity

Lxran:
Also added cover image to GHPG


Git-GoR:
Seems like the errors are not due to the files changed in this PR, but caused by the `chore: move subclasses to root` 7febdab709441e6039b4c23f1b6de91ae009bdc1 commit. 
Before, `subclass`es that were inside a class inherited the class' source, which I guess is not the case anymore.

TheGiddyLimit (approved):

PR #995, opened 2021-06-09
Uploaded images for future conversion

ineedanamepls:
- Uploaded images for "Stibbles Codex of Companions" conversion


ineedanamepls:
Actually have to make them .webp first

PR #1056, opened 2021-08-08
Hotfix for brew with wrong property on a weapon

Foxfire94:
One of the brews needed a quick fix to patch an error that was caught by a user.


TheGiddyLimit (approved):
Please grab properly-formatted copies from the repo before making changes; this diff would be ~10 lines instead of ~16,000

PR #1062, opened 2021-08-12
Add unofficial ru translation for 5e spells.

vvechkanov:


TheGiddyLimit (changes requested):
Nice work, although I'm not sure this repo is the best place for storing translation files.

Either way, it looks like you've only replaced the first line of `entries`, and have left partial English versions after the Russian translation. What's the reason for this?

PR #1067, opened 2021-08-22
Update laserllama; The Savant.json

Assasinine1:
Updated to 3.5.1


TheGiddyLimit:
Looks like you've broken the JSON file. Run it though [jsonlint](https://jsonlint.com/)!

TheGiddyLimit:
It also looks like you've also dropped the `"subclasses"` array--what was the thinking here?

TheGiddyLimit (changes requested):
(as above)

PR #1079, opened 2021-09-08
Split "D&D Wiki" homebrew into two files, with more descriptive names

jpcranford:


creature/D&D Wiki; Rabbit and Squirrel.json
```diff
@@ -0,0 +1,160 @@
...
+                    ]
+                }
+            ],
+            "damageTags": [
+                "P"
+            ],
+            "miscTags": [
+                "MW"
+            ],
+            "uniqueId": "7cbe3f0531e940888e7903fbb6032ceb",
```
Git-GoR, line 64: Be aware that `uniqueId` gets created automatically when the brew is loaded, and is not needed in the homebrew file.

Unsure if it would cause any issues.

collection/D&D Wiki; Killer Rabbit.json
```diff
@@ -0,0 +1,220 @@
...
+            "damageTags": [
+                "P"
+            ],
+            "miscTags": [
+                "MW"
+            ],
+            "conditionInflict": [
+                "grappled"
+            ],
+            "uniqueId": "5454ae4290873a69a334bc0a43f36cba",
```
Git-GoR, line 167: Same thing as above

creature/D&D Wiki; Rabbit and Squirrel.json
TheGiddyLimit: Yeah, those shouldn't be in there. They'll get clobbered as it gets loaded, either way.

TheGiddyLimit (approved):

PR #1088, opened 2021-09-12
Update and rename Dark Matter

englishkiwi:
Updated the various meta fields, and added an extension to the file. Oops!

### DIFF
--- a/collection/Mage Hand Press; Dark Matter - Partial.json
+++ b/collection/Mage Hand Press; Dark Matter - Partial.json
@@ -5,21 +5,26 @@
 				"json": "DarkMatter",
 				"abbreviation": "DaM",
 				"full": "Dark Matter",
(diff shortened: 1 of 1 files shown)


englishkiwi:
Just a note that this has failed on the npm test for the following:

```
{
    "keyword": "type",
    "dataPath": "._meta.senses['thermalsight']",
    "schemaPath": "#/properties/_meta/properties/senses/patternProperties/.*/type",
    "params": {
      "type": "string"
    },
    "message": "should be string"
  }
```

It's asking for a string, but the file works perfectly when I upload it to homebrew manager. When I change it to just a string the end result on 5etools is every letter of the string on a new line (screenshot attached). Is there potentially a problem with the schema?

Screen Shot 2021-09-13 at 8 53 36 AM

PR #1094, opened 2021-09-24
Update Mage Hand Press; Dark Matter - Partial.json

englishkiwi:
- Added all vehicles, feats, backgrounds, ship weapons, ship systems.
- Added new Construct Graft, System and Ship Weapon item types.


englishkiwi:
Failed with similar issue to last pull request, this time with the custom Skills. The file works fine when uploaded to Homebrew Manager. Happy to make any necessary changes if someone can point me in the right direction.

TheGiddyLimit (changes requested):
Any reason you've nested `"entries"` objects instead of a plain array of strings? That's the only thing it's getting caught on

englishkiwi:
Oh strange, I was sure last I tried it the array of strings didn't work, but I've just tested it and it did. Sorry about that, I've  added another commit with those changes made.

PR #1100, opened 2021-09-30
Update Dragon Knight class to 8.1

NutritiousCookie:
minor changes, reformatted JSON for readability


TheGiddyLimit (approved):
Please _do not_ reformat according to your own specific idea of readability, as this breaks consistency with the rest of the repo.

Regardless, such changed will be auto-formatted away anyway, so all this does is create junk diffs

PR #1130, opened 2021-10-29
Replace/update LevelOfExhaustion; Circle of the Sands.json

Louswastaken:
Original file results in spaces missing in the search query for "Get Homebrew". This resolves the issue.


TheGiddyLimit (approved):
Still missing a space after the semicolon, but, I'll fix that 👍

PR #1142, opened 2021-11-18
Fix minor schema issues in FFXIV5E

Lxran:


Lxran:
Messed up the branches, see #1143

PR #1156, opened 2021-11-30
Added Mudborn Homebrew

cadowtin:


TheGiddyLimit (changes requested):
Looks like this needs to be run through https://www.jsonlint.com/

cadowtin:
Perhaps now? Removed the extra comma that was causing issues

PR #1167, opened 2021-12-13
Edited pure Strength Die from the Empowered Class

Genocide-Games:
it's supposed to scale, how did they miss that?


Genocide-Games:
i'm an idiot

TheGiddyLimit (approved):

PR #1174, opened 2021-12-21
MME3 Images

Spappz:
Adding images early for testing.

These images are ripped from a FG .mod and are lower-quality than can be ripped from the PDF. I am _considering_ replacing them at a later date... but if that will sully the git history you're welcome to close this without merging and I'll do it all at once later.


TheGiddyLimit:
👍 I'd rather not have a bunch of extra junk added to the history if it can be avoided

PR #1196, opened 2022-01-10
Charactermancer fixes

cadowtin:
Updated Spells specifically adding a the spellcasting modifier, and the caster type to half caster so that spells generate in charactermancer.
Added upgrades as optional features that get assigned at their appropriate levels


TheGiddyLimit (approved):
👏

PR #1214, opened 2022-01-24
Firearms of the Realm

cadowtin:


cadowtin:
Only a partial implementation, but enough to be usable

TheGiddyLimit (approved):

PR #1220, opened 2022-01-27
Add "Tome of Mystical Tattoos"

0rganics:
(apologies if this takes even more commits, I can't figure out how to run the schema or test script locally ._. )


TheGiddyLimit:
No worries. If you want to keep things clean, you can install and use git (rather than doing things via the website), where there's an option to "amend" your last commit (`git commit --amend` in the cli). You can then `--force` a push to overwrite the previous version of your commit

PR #1227, opened 2022-02-03
Add Martial Power book

Git-GoR:
By badooga

It contains the class features (and links to them in the book), but I did not create copies of every class cause 😩😪
This should make them searchable on the site/importable with Plut though, which is good enough.


Git-GoR:
Clicked the "merge master with branch" button on the PR by accident, why does that exist 😒

Git-GoR:
Actually nvm, there doesn't seem to be any conflicting changes. Good to go 👍

TheGiddyLimit:
the "merge master with branch" button exists because I turned it on 😏

PR #1265, opened 2022-03-02
Fixing Tables

cadowtin:


cadowtin:
Unsure why it isn't compiling 
Also book has significant misc fixes (improper formatting, missing text, etc.)

cadowtin:
TODO:

- Make sure all Companions have all their text
- Make sure all Companions have their text formatted correctly
- Add Token Images for every companion

TheGiddyLimit (approved):

PR #1283, opened 2022-03-14
Add files via upload

Das-Moko:
Initial commit of Runekeeper class & magic system


revilowaldow:
## Biggest problem:

Almost everywhere (125/129) you've used `otherSummary` you need to either;

-  Include an `entrySummary` value next to `entry` for use in the lists.
OR
- Change `otherSummary` to `other`


##  Additionally you have a few other small fixes to make:
#### Line 139:
Replace `"shield"` with `"{@item shield|phb|shields}"`
#### Line 3821:
Replace `"name"` with `"caption"`
#### Line 9760:
Change 
```json
"size": "Medium or larger",
```
for
```json
"size": "V",
"sizeNote": "Medium or larger",
```
#### Line 9762:
Remove
```json
"alignment": null,
```

TheGiddyLimit (approved):

PR #1294, opened 2022-03-24
Fix tag issues introduced by cleaning step

revilowaldow:


revilowaldow:
This seems to have failed on somebody else's brew already in the repo? https://github.com/TheGiddyLimit/homebrew/pull/1294/checks#step:5:815

PR #1340, opened 2022-05-07
Foundry Support alongside some rewording/ordering.

Maalunar:


Maalunar:
If I understand that correctly, it appears that the schema doesn't accept any FoundryVTT line beside foundryImg for feats. So it reject foundryEffects and foundryData.

Maalunar:
Thought myself smart. Had given the spell selection from the Ritual Caster talent 0 slot instead of the 1 the regular feat had in 5etools. With 0 slot it won't give any in FoundryVTT, perfect for ritual caster. But 0 ain't a valid, oh well. It works at least now.

PR #1348, opened 2022-05-12
Update devikyn; Variant Sorcerer.json

MrVauxs:
Fix feat tags


MrVauxs:
The error has nothing to do with the changes I've done lmao

PR #1352, opened 2022-05-21
Create teste

Danilocpm:


revilowaldow:
Definitely in the wrong folder, unsure if the folder locations check would catch this? 
Guess we'll find out when it's run.

TheGiddyLimit (changes requested):
I doubt it. Either way, this is mince. Please move to `spells/Bonfire Tales; Bonfire Tales Tradução de Magias.json`

PR #1369, opened 2022-06-05
The Tinker Class

entvidya:
A homebrew revision of the artificer


entvidya:
I messed up, will fix it

PR #1370, opened 2022-06-05
Add The Tinker Class, an Artificer revision

entvidya:
Tested on both 5eTools website and Foundry, should be working great.


entvidya:
Nevermind, I think I see what it didn't like. The schema didn't complain and neither did Foundry, so I thought it would be fine. Changed the Lesser Restoration from Chemist subclass to 1 use per day for now, will need to be manually adjusted when importing.

TheGiddyLimit (approved):

PR #1375, opened 2022-06-12
Update Blood Hunter 2020

EldritchTranslator:
Updated BH2020 with the edits from Feb 2022 on DDB


revilowaldow:
This should probably be a new file called BH2022...

TheGiddyLimit:
Should it? What's the justification for/against? Thoughts as the converter, @MrVauxs ?

EldritchTranslator:
Considering it's really mostly an update to clarify some abilities and updating them with new warlock patrons, I didn't think of making a separate file, as with like, the original bloodhunter. It looks like I rewrote a lot, but they rewrote basically the same things.

MrVauxs:
> Should it? What's the justification for/against? Thoughts as the converter, @MrVauxs ?

image


I specifically **didn't** update it because it wasn't an official update, but one made by D&DB exclusively for D&DB. 
It should be it's own separate file.

TheGiddyLimit:
So noted. We'll go with a second copy. `... (2022)`, then. Closing for now; @EldritchTranslator feel free to either update this PR and it can be re-opened/merged, or create a new one. Or if you want me to do it, let me know

EldritchTranslator:
Ah, forgot about this. I'll do another PR in a bit, sorry

PR #1399, opened 2022-07-20
Add Sleuth background

jpcranford:
Didn't know how to create the last modified hash, so I deleted it. I presume it auto-generates?


TheGiddyLimit (approved):

TheGiddyLimit:
It does, yes. It's based off the file contents, too, so you don't need to delete it; it'll automatically get updated when I next build the repo

PR #1415, opened 2022-08-01
Update BeardedJB; Cosmic Cleric Domain.json

JavaShipped:
Attempting to fix an undefined error `TypeError: Cannot read properties of undefined (reading 'subclassShortName')` by fixing the `"subclassSource"` which was still defaulted to `"source": "TCE"` instead of the new `"source": "CosmicDomain"` in most cases.


JavaShipped:
I just tested by uploading the file - doesn't work still, needs further investigation

PR #1426, opened 2022-08-12
Update DianaDM; A Series of Stars.json

Voxeldoodle:
Updated the ASoS module as published by u/Liusediana at https://www.gmbinder.com/share/-MovsNJZ1mFAks1B2Gpw

This includes:
- the Relativity Astromancer subclass
- all the remaining subclasses for:
     - Bard
     - Druid
     - Fighter
     - Monk
     - Paladin
     - Ranger
     - Sorcerer
     - Warlock (2 Additional Eldritch Invocations)
     - Wizard
- The new race: Xel'na, along with its 2 subraces:
      - Consumed
      - Void Touched
- the Farsoul Dragonborn subrace


Voxeldoodle:
A couple of things to note:
I did this update by hand, being the first ever to make, so even though I tried to be as thorough as possible, something might have escaped me.
In particular, I just skimmed through Items and Monsters, changing only the page reference.

For the images urls, I used the ones form gmbinder; I don't know if it's ok or if something else needs to be done.
Specifically the ones for the Xel'na race and the one for the book cover.

TheGiddyLimit (changes requested):
Needs fixing up--I'd recommend getting set up with something that can check your file against the schemas. #brew-conversion in our Discord should have more info.

Voxeldoodle:
Fixed the error on the fly, and therefore I haven't checked it properly. It should work though, because it was just the eldritch invocations duplicated (they were already listed as optional feature, but I left them in the subclass description as well).  

If you want you could try rerun the checks, otherwise I will check it properly tomorrow if I can.

TheGiddyLimit (approved):

PR #1427, opened 2022-08-13
Update teqqqie; Weird Flavor Creatures.json

teqqqie:
Updated Corpse Masker description and tweaked statistics


teqqqie:
I've looked at the error and I don't understand why the check failed

revilowaldow:
Giddy has updated the schema, but not updated the contents of the repo to pass it yet. This means that everyone's requests will fail on a file that's already in the repo until he takes a look at it and migrates everything to the new system. 

(Specifically optional feature types are always an array now, not an array or a string)

TheGiddyLimit (approved):

PR #1440, opened 2022-08-23
Update MCDM Productions; Strongholds and Followers.json

Spappz:
goodbye to the greatest feat of html injection in this repo 😔


revilowaldow:
yeet

TheGiddyLimit (changes requested):
`$rgb-name-red` isn't a valid color, it only "works" because it gets regex'd down to the valid hex characters `baeed`, although this isn't a valid color either. If you're seeing it be the right color, this is because the title gets colored that way anyway

revilowaldow:
It overwrites the blue from the variant rule tag which had previously overwritten the correct header colour, and then works correctly with the night mode switching. So something a little more complex is going on. If you asked me to chose between the abomination that was there before and a dodgy colour tag I know what I'd pick...

TheGiddyLimit:
> It overwrites the blue from the variant rule tag which had previously overwritten the correct header colour

on second thought this doesn't make any sense--links, even those in headers, are _supposed_ to be blue, as this is the only way of identifying that something is a link. So either way, pls fix

Spappz:
> links ... are supposed to be blue

I kinda liked it when the link was 'subtle', and you only notice it by hovering. The blue is a bit too obvious—good for small text, not so much for headers 😩

TheGiddyLimit:
then remove it from the header, and put the link in the `entries`!

Either way, looks like you've got some merge conflicts too 🙃

TheGiddyLimit (approved):

Spappz:
> remove it from the header

The original statblocks in the book have a special header section (like Actions and Reactions) and I'm just trying to emulate that. Putting it in the `entries` would mean adding random text on the end 🥲

Also I have absolutely no idea what the merge conflicts are now so I'm just going to nuke this branch and PR again 😎

PR #1455, opened 2022-09-02
Warlock Patron Ghost Train

cadowtin:


hakr14:
The "source" argument of the objects needs to be the json identifier of the homebrew source, not the abbreviation.

TheGiddyLimit (approved):

PR #1495, opened 2022-10-01
spells from Heliana's guide to monster hunting

Alexandre-Pe:
Spells from Heliana's guide to monster hunting along with roll20 and foundry integration when possible.


revilowaldow:
These should all be in one file in the collection folde

PR #1504, opened 2022-10-15
Add Water Domain

calebmeyer:
Everything appears to be working now: 
image

### DIFF
--- a/subclass/GM Binder; Water Domain.json
+++ b/subclass/GM Binder; Water Domain.json
@@ -105,15 +105,15 @@
 						],
 						[
 							"5th",
-							"{@spell tidal wave}, {@spell water breathing}"
+							"{@spell tidal wave|xge}, {@spell water breathing}"
 						],
 						[
 							"7th",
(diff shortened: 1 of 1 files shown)


subclass/GM Binder; Water Domain.json
```diff
@@ -0,0 +1,204 @@
...
+				],
+				"convertedBy": [
+					"calebmeyer"
+				],
+				"version": "3.2"
+			}
+		],
+		"dateAdded": 1234,
+		"dateLastModified": 1234,
+		"_dateLastModifiedHash": "todo"
```
calebmeyer, line 21: I wasn't sure how the meta updater works, so I put some garbage values here. I can remove these lines if needed.

TheGiddyLimit: you can remove the `_dateLastModifiedHash`; it's auto-generated when I index things.
As for the other two, you can use e.g. https://www.epochconverter.com/ and grab the timestamp at the top

PR #1558, opened 2022-12-09
A bundle of subclasses for various classes

Gelidity:
1 update to an existing subclass of mine
2 new subclasses of mine
1 new subclass by FrenchTech16 that I converted


TheGiddyLimit (changes requested):
put them in the `subclass/` directory, please

TheGiddyLimit (approved):

PR #1575, opened 2022-12-19
LaserLlama's Alt Ranger Update 3.7.1 -> 3.8.1

Ignimortis:
An update to LaserLlama's Alternate Ranger (plus Expanded) homebrew. Includes three new subclasses and alterations to old subclasses, all according to the LL's PDFs on GMBinder.


revilowaldow:
File should be renamed `LaserLlama; Alternate Ranger.json` to match the existing file and needs to be placed into the `class` folder

PR #1589, opened 2022-12-30
Update for Blake and Kevin (Blurhy); The Armiger.json

kevingorduno:
Realized the class version on the metadata is incorrect. It's V2.4 and not v2.1. Also fixed one typo and added an @ tag to the Armiger's Armory feature within the Manifested Armory feature just for convenience.

### DIFF
--- a/class/Blake and Kevin (Blurhy); The Armiger.json
+++ b/class/Blake and Kevin (Blurhy); The Armiger.json
@@ -5,7 +5,7 @@
 				"json": "TheArmigerClass",
 				"abbreviation": "AMGR",
 				"full": "Armiger",
-				"version": "2.1",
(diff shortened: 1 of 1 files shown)


kevingorduno:
Friendly notice that we will be releasing v2.5 soon which is when we'll announce the 5e.tools integration (and the new PDF we're making for the class - woohoo!). So, there will be another update soon. This is just a heads up.

But, we wanted the class up early (the v2.4 version) to test out integration and everything... like a soft launch of the 5e.tools integration. 😅 Just an FYI.

TheGiddyLimit:
@kevingorduno if you need something done in a timely manner, do ping me on Discord. I check this repo ~once per week, on average

PR #1602, opened 2023-01-17
Fixed Errors on Llaserlama's Alt Rogue

cadowtin:
Fixed errors with subclass feature naming
added formatting for die rolls
Fixed some text for skinchanger

### DIFF
--- a/class/LaserLlama; Alternate Rogue.json
+++ b/class/LaserLlama; Alternate Rogue.json
@@ -1676,7 +1676,7 @@
 			"subclassFeatures": [
 				"Justicar|Alternate Rogue|LLARo|Justicar|LLARoE|3|LLARoE",
 				"Consecrated Strikes|Alternate Rogue|LLARo|Justicar|LLARoE|7|LLARoE",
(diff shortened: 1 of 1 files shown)


CrazierIvan:
This fix is better than my fix, so I'm limiting my pull to warlord only.

class/LaserLlama; Alternate Rogue.json
```diff
@@ -6176,7 +6177,7 @@
 					"type": "table",
 					"caption": "Bewildering Blow",
 					"colLabels": [
-						"1d10",
```
TheGiddyLimit, line 6179: fwiw both the dice header _and_ the numbers in the first column should get auto-detected; you shouldn't need to specify `@dice` in the header + `cell`s 🤔 Did it not work without these?

cadowtin: Oh, it probably did  i don't think I checked. I was more following the format of the Render Demo (I reference it occasionally when writing up homebrew stuff)

PR #1632, opened 2023-02-14
Subclasses from Crown of the Oathbreaker

CardboardandVinyl:
36 subclasses (3 for each SRD class) from the Elderbrain Crown of the Oathbreaker campaign.


TheGiddyLimit (approved):

CardboardandVinyl:
If I had to make changes to the json because not all the content would load (I think I had issues with headers and refSubclassFeature) do I need to submit a new pull request or can I edit the version you've merged?  If it is a new PR then do I just keep the same name and it replaces the original once merged?  Thanks for your help

PR #1633, opened 2023-02-14
Add files via upload

DrGoo282:
A conversion of Kibbletasty's Inventor class, uses the spell compendium reference.


revilowaldow:
Shouldn;t this file update https://github.com/TheGiddyLimit/homebrew/blob/master/class/KibblesTasty%3B%20Artificer%20(Revised).json
Which is the Inventor before the name was changed.
### Original:
image 

### Changed to:


image

PR #1636, opened 2023-02-17
Updated Elderbrain; CotO Subclasses

CardboardandVinyl:
Fixed multitude of formatting issues, added more appropriate headers, added refSubclassFeatures.


CardboardandVinyl:
I think the lint test failed because there is an existing file with the same name?  

[  FILES_ROOT] Testing for unwanted JSON files in root dir...
Unwanted JSON file in root directory: Elderbrain; Crown of the Oathbreaker Subclass Compendium.json

GoR had told me to use the same file name as the previously merged file. Is this what is causing that issue?

PR #1696, opened 2023-04-24
Miz's magical items

Mizeov:
These items come from a variety of sources some custom made, some from the DMG. I am testing to see if I can give them a price and format them correctly so that they can be pulled into foundry Vtt. Future items will be created here as well.

first tested with a potion of greater healing to see if I could give a price and a healing amount. Soon to add custom Blunderbuss item as well as other custom items my players may wish

Note: Mods or whoever reads this I have no idea what I'm doing so hopefully i was able to MacGyver it correctly. The newest file that has healing at 4d4+4 is the correct amount for a potion of greater healing. I don't know if it simply updates the file I uploaded a few minutes ago which i tried to dele


TheGiddyLimit (changes requested):
🤔 I see you've posted in the #brew-conversion channel in our Discord and gotten some help, is this the most recent version of the file?

If so, `"Value": "20000"` should be `"value": 20000` (keys always:tm: start with a lowercase letter, and numbers _generally_ shouldn't be in quotes), and `"Healing"` isn't a valid property.

The file should also be in the `item/` folder.

Are you intending to use this in Roll20, or in Foundry? (Or neither?)

Mizeov:
The people in the discord were able to help me do what i was trying to do so I dont need github anymore. I'll upload new files in the future if I need though I do not know how to delete this. thank you!

PR #1701, opened 2023-04-30
first release

Fratepietro87:
add dante inferno creature


Fratepietro87:
let me know if is ok or i have to performe any change.
Regards

PR #1728, opened 2023-05-13
second release

Fratepietro87:


Fratepietro87:
hi, what's was wrong?
i just only want to update my creature file with new creatures.
let me know plse
regards
Giorgio

TheGiddyLimit:
If you take a look in the ["Checks" tab](https://github.com/TheGiddyLimit/homebrew/pull/1728/checks), you can see the exact error message--the problem is that your source is used in two different files. A source can only be defined in one file.

To resolve the issue, you can:

- Rename your source (just the "json" part) and all the `"source": "..."` fields on your data to match, or,
- Move all your stuff into one file and stick it in the `collection/` directory

revilowaldow:
image

Think it's just a file name change the content is similar just getting up issued

Fratepietro87:
realy tnx guy, i'm going to fix it.
regards

PR #1759, opened 2023-06-17
Add files via upload

SLankoff:
If appropriate here, I made this mostly for me.

### DIFF
--- a/class/Mage Hand Press; Vagabond.json
+++ b/class/Mage Hand Press; Vagabond.json
@@ -1,0 +1,2558 @@
+{  "_meta": {
+    "sources": [
+        {
+            "json": "MHP:VB",
+            "abbreviation": "MHP:VB",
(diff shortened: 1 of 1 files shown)


TheGiddyLimit (approved):

TheGiddyLimit:
a perfectly valid use case

TheGiddyLimit:
@SLankoff your Foundry description HTML for the class didn't make it through unscathed--there's an automated HTML scrubber (which in this case is probably over-tuned, but that's a tricky subject) which ate it. Was there any reason you specified the description instead of using the default generated one?

SLankoff:
Oh gosh i hope this reply gets to you...

Yes, though it was super super minor. In foundry the reference to the class
feature in the fluff text links to 5etools instead of the feature directly,
so I linked it using foundry item linking.

PR #1779, opened 2023-07-01
No changes

hakr14:
just appeasing the git gods


TheGiddyLimit:
what's the supposed to be? Merging would add a bunch of junk history (adding then removing image?) to the main repo, which doesn't seem useful

PR #1808, opened 2023-08-09
Flee mortals

Billiam9420:
Let me know what I messed up lol I was half dead for a lot of this and some of the formatting changed as I went along


Billiam9420:
Schema has issues with the following three things:

1. "resources" says that "name" is not allowed, but I'm pretty sure it should be required.
2. "resist" (and possibly "immune" and "vulnerable") does not allow custom input, it requires listing from the normal damage types.
3. "spellcasting" requires "type" but it seems the only option allowed is "spellcasting".

The site itself has no issues with these, but the schema does.

TheGiddyLimit:
A thought: it would be nice to have `"environment"` set on each creature which is from a specific environment (`"cave"`, etc.)

Billiam9420:
Everything should be good to go now!

TheGiddyLimit (approved):

PR #1818, opened 2023-08-24
Fix @ tags

Spappz:
One flagged by a human, and one by an Excellent™ new [script](https://github.com/Spappz/Tools45eTools/blob/master/tag-checker.ps1) I have written.


revilowaldow:
Considered making this fancy new script part of the existing test suite?
https://github.com/TheGiddyLimit/5etools-utils/blob/497eb1b5d6515491c945ace8bc31bd471cf32c82/lib/TestData.js#L86-L108

TheGiddyLimit (approved):

PR #1823, opened 2023-08-30
Repair - GuenMaster; Way of the Stars

AmbyNavy:
Repaired to function as well as I can. The class ability aren't showing properly inside of Charactermancer and I can't see why, but everything imports, loads and works now.


TheGiddyLimit (approved):
long time no see!

PR #1839, opened 2023-09-16
Update Dungeon Dad; Dungeon Dad Statblocks.json

Genocide-Games:
Updated satblocks


revilowaldow:
@TheGiddyLimit did that double \\ for the cleaner get put back in? This looks like the cleaner had replaced the quotes and broken it during the tests?

Genocide-Games:
not %100 sure what your talking about, I just replaced instances of [. "] with [."] thinking the space between the period and the quotation mark screwed things up.

revilowaldow:
If you have a `"` in a string, it should be preceded by an escape character `\` to let the json file know that it isn't the end of the string. 
There's a cleaner that replaces squiggly quotes `”“` with `\"` but I think it's not adding the \ at the moment so it actually breaks the file while trying to test it.

TheGiddyLimit:
it _should_ be fixed, but if not, I'll clean it up post-merge

TheGiddyLimit (approved):

PR #1844, opened 2023-09-24
Updates Alternate Fighter and Expanded to 3.1.0

CrazierIvan:


CrazierIvan:
Okay, I think we're done.

CrazierIvan:
Now we're actually done.

TheGiddyLimit (approved):

PR #1869, opened 2023-10-10
Fix footnotes in Flee Mortals

ebullient:
Use the footnoteEntries attribute in spellcasting to explain symbols for casting time

### DIFF
--- a/creature/MCDM Productions; Flee, Mortals!.json
+++ b/creature/MCDM Productions; Flee, Mortals!.json
@@ -11668,16 +11668,19 @@
 						"In addition to any other spells in this stat block, the channeler can cast the following spells, using Charisma as the spellcasting ability (spell save {@dc 15}):"
 					],
 					"will": [
-						"{@spell light}{@footnote {@sup A}|Casting time: 1 action}",
-						"{@spell message}{@footnote {@sup A}|Casting time: 1 action}"
+						"{@spell light}{@sup A}",
+						"{@spell message}{@sup A}"
 					],
 					"daily": {
 						"1e": [
-							"{@spell detect magic}{@footnote {@sup A}|Casting time: 1 action}",
-							"{@spell disguise self}{@footnote {@sup A}|Casting time: 1 action}",
-							"{@spell identify}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}"
+							"{@spell detect magic}{@sup A}",
+							"{@spell disguise self}{@sup A}",
+							"{@spell identify}{@sup +}"
 						]
 					},
+					"footerEntries": [
+						"Casting time: {@sup A} = 1 action; {@sup +} = Longer than 1 action (see spell description)"
+					],
 					"ability": "cha"
 				}
 			],
@@ -17158,12 +17161,15 @@
 					],
 					"daily": {
 						"3e": [
-							"{@spell commune}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}",
-							"{@spell commune with nature}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}",
-							"{@spell legend lore}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}",
-							"{@spell scrying}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}"
+							"{@spell commune}{@sup +}",
+							"{@spell commune with nature}{@sup +}",
+							"{@spell legend lore}{@sup +}",
+							"{@spell scrying}{@sup +}"
 						]
 					},
+					"footerEntries": [
+						"Casting time: {@sup +} = Longer than 1 action (see spell description)"
+					],
 					"ability": "wis"
 				}
 			],
@@ -22796,20 +22802,23 @@
 						"In addition to any other spells in this stat block, the hag can cast the following spells, using Charisma as the spellcasting ability (spell save {@dc 14}):"
 					],
 					"will": [
-						"{@spell alter self} {@footnote {@sup A}|Casting time: 1 action}",
-						"{@spell dancing lights} {@footnote {@sup A}|Casting time: 1 action}",
-						"{@spell minor illusion} {@footnote {@sup A}|Casting time: 1 action}"
+						"{@spell alter self}{@sup A}",
+						"{@spell dancing lights}{@sup A}",
+						"{@spell minor illusion}{@sup A}"
 					],
 					"daily": {
 						"1": [
-							"{@spell scrying} {@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}"
+							"{@spell scrying}{@sup +}"
 						],
 						"3e": [
-							"{@spell animal messenger} {@footnote {@sup A}|Casting time: 1 action}",
-							"{@spell legend lore} {@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}",
-							"{@spell speak with animals} {@footnote {@sup A}|Casting time: 1 action}"
+							"{@spell animal messenger}{@sup A}",
+							"{@spell legend lore}{@sup +}",
+							"{@spell speak with animals}{@sup A}"
 						]
 					},
+					"footerEntries": [
+						"Casting time: {@sup A} = 1 action; {@sup +} = Longer than 1 action (see spell description)"
+					],
 					"ability": "cha"
 				}
 			],
@@ -22976,20 +22985,23 @@
 						"In addition to any other spells in this stat block, the hag can cast the following spells, using Charisma as the spellcasting ability (spell save {@dc 15}):"
 					],
 					"will": [
-						"{@spell alter self}{@footnote {@sup A}|Casting time: 1 action}",
-						"{@spell detect magic}{@footnote {@sup A}|Casting time: 1 action}",
-						"{@spell pass without trace}{@footnote {@sup A}|Casting time: 1 action}"
+						"{@spell alter self}{@sup A}",
+						"{@spell detect magic}{@sup A}",
+						"{@spell pass without trace}{@sup A}"
 					],
 					"daily": {
 						"1": [
-							"{@spell scrying}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}"
+							"{@spell scrying}{@sup +}"
 						],
 						"3e": [
-							"{@spell dream}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}",
-							"{@spell legend lore}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}",
-							"{@spell mirage arcane}{@footnote {@sup +}|Casting time: Longer than 1 action (see spell description)}"
+							"{@spell dream}{@sup +}",
+							"{@spell legend lore}{@sup +}",
+							"{@spell mirage arcane}{@sup +}"
 						]
 					},
+					"footerEntries": [
+						"Casting time: {@sup A} = 1 action; {@sup +} = Longer than 1 action (see spell description)"
+					],
 					"ability": "cha"
 				}
 			],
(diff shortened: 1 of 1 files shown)


TheGiddyLimit:
"Fix" is a strong word, what, in particular, is wrong with the current version? The `@footnote` usage was a specific choice I made, and even added to the [creature converter](https://github.com/5etools-mirror-1/5etools-mirror-1.github.io/blob/master/js/converterutils-creature.js#L1470)!

revilowaldow:
For context, ebullient makes the 5eTools (& PF2e) to Obsidian Markdown tool.
https://github.com/ebullient/ttrpg-convert-cli

TheGiddyLimit:
> For context

ty I did remember this time

---

> If you hate this plan

"Hate" is another strong word, but I _do_ like my nice hover-able superscript letters. That said, on closer inspection, the `@footnote` tag is a bit broken; it relies on some in-memory hackery and so doesn't work when e.g. importing a creature via Plutonium and then refreshing the page.

I think my path of least resistance here would be to make it into a loadable entity type i.e.

```json
"footnoteEntry": [
  {
    "name": "Spell Action Note",
    "source": "FleeMortals",
    "page": 123,
    "entries": [ ... ]
  },
  ...
]
```

and then `@footnote` would work like any other entity-loading tag (`@spell`, `@creature`, etc.) and so we'd have something of the form:

```json
"{@spell detect magic}{@footnote {@sup A}|Spell Action Note|FleeMortals}, {@spell disguise self}{@footnote {@sup A}|Spell Action Note|FleeMortals}"
```

This would fix the Plutonium jank (For Free:tm: by way of the existing hover-able tag handling paths), and de-dupe all the footnote strings. I imagine you could then decide how you wanted to deal with linked `@footnote`s in an object; you could scan all the strings and pull out the tags and render them in a bit at the end, or, ... .

Thoughts?

ebullient:
> > If you hate this plan
> 
> "Hate" is another strong word, but I _do_ like my nice hover-able superscript letters. 

=) Fair enough. The hover is definitely nice for the web UI, it just doesn't work well in Obsidian or straight markdown.

> I think my path of least resistance here would be to make it into a loadable entity type i.e.
> 
> ```json
> "footnoteEntry": [
>
>     "name": "Spell Action Note",
>     "source": "FleeMortals",
>     "page": 123,
>     "entries": [ ... ]
>   },
>   ...
> ]
> ```
> 
> and then `@footnote` would work like any other entity-loading tag (`@spell`, `@creature`, etc.) and so we'd have something of the form:
>
> ```json
> 
> "{@spell detect magic}{@footnote {@sup A}|Spell Action Note|FleeMortals}, {@spell disguise self}{@footnote {@sup A}|Spell Action Note|FleeMortals}"
> 
> ```

Does this break existing footnotes:

- `{@footnote directly in text|This is primarily for homebrew purposes, as the official texts (so far) avoid using footnotes}`  
- `{@footnote optional reference information|This is the footnote. References are free text.|Footnote 1, page 20}`

For example: 

"Rorreth (Red Wizard) has the following {@footnote wizard spells prepared|Replaced non-existent {@b fire blast} spell from original text with {@b fire bolt}}:"

Feels like you're changing semantics here. Footnotes are already used in other homebrew, and that middle section is usually real text (even in the case where the third segment is present).

Can we do something like `{@footnoteRef key|FootNote Name|source}`, where that key is the known/fixed superscipt value (no `{@sup..}`)?

revilowaldow:
There's over 500 uses of `{@footnote` in the homebrew repo, as long as you have a plan to migrate them I don't have an objection.
image

See also `{@homebrew` tags if you're hoping to factor this similar tag behaviour out.
image

ebullient:
With the proposed changes, I generate something like this: 

image

ebullient:
Given the changes above, I would generate something like this (appearance varies greatly by theme): 

image

I agree that the lack of hover over the little A and the + is sad, but there is a not a great way for me to add that using normal footnote processing (as you realized, I think).

I can make a `footnoteRef` work. I think there is a similar situation (well-known/reused footnote key) in tables (which also have footerSo maybe there is another noun that would be better (footerRef), that is implicitly more localized..  entries of some kind). So maybe there is another noun that would be better to indicate superscript/linked references to localized/nearby footer content. (as another thought)

TheGiddyLimit:
> Does this break existing footnotes

> a plan to migrate

yes and yes, they're trivial enough to migrate

> Can we do something like `{@footnoteRef key|FootNote Name|source}` instead

apologies; smashed out meaningless junk before going to bed. What I meant was: `{@sup {@footnote Spell Action Note|FleeMortals|A}}`

> having both `@footnote` and `@footnoteRef`

that's a definite _hate_. I'm not maintaining two systems for the same thing, especially for something as uninteresting as footnotes

> pain incarnate

is it really, though? Make one/or an IDE template of your choice with a generic name (say `"name": "Footnote p#"\n"source": "..."`), and it's about as quick as the current system, no..? Presumably one has chop up text currently to inline a footnote so I don't see this as a huge workflow change

> `@homebrew` tags

those are likely broken too, but iirc the implementations are separate, so we can table that

---

really what I'm asking is, "is this any different to a `@skill`/`@sense` tag." From an implementation standpoint, it's a firm "no," and I'm hesitant (aka it ain't happening) on making things more complex by adding e.g. block-local references. If `@footnoteRef` is workable then that can be the new `@footnote`!

revilowaldow:
>  it's about as quick as the current system, no..?

I agree, it's fine in terms of the actual technical challenge of making an entry.

The pain is that in a brew with 80 unique footnotes I now need to come up with 80 source-unique `"name"`s for what is effectively an anecdotal sentence.
When you migrate these, are you going to come up with sensible names or automate them 1 through 500? Using the nearest entry header might help but multiple footnotes within an entry are non-unique and the entry header could well be completely disconnected from the anecdote contained in the footnote.

I thought the point of footerEntries on a Spellcasting entry was that this is how wizards and inspired brew sources laid out the notes for spell lists?

Changing the whole footnote system which is used very differently in other parts of the site to enable this specific brew feels strange. Though I understand the wish to factor out non-standard behaviour.

TheGiddyLimit:
I expect a "footnote" to be something like this:

image

> I now need to come up with 80 source-unique names

so the "uniquely named entity version" of it would just be this:

```json
"footnoteEntry": [
  {
    "name": "Footnote 17",
    "source": "MySource",
    "page": 250,
    "entries": [ ... ]
  },
  {
    "name": "Footnote 18",
    "source": "MySource",
    "page": 250,
    "entries": [ ... ]
  },
  ...
]
```

or for a book which doesn't uniquely number their footnotes:

```json
"footnoteEntry": [
  {
    "name": "Footnote 1 p250",
    "source": "MySource",
    "page": 250,
    "entries": [ ... ]
  },
  {
    "name": "Footnote 2 p250",
    "source": "MySource",
    "page": 250,
    "entries": [ ... ]
  },
  ...
]
```

> footerEntries on a Spellcasting

agreed..? Not sure I follow, this is not about changing `footerEntries` (no plans there)

> Changing the whole footnote system which is used very differently in other parts of the site to enable this specific brew feels strange. Though I understand the wish to factor out non-standard behaviour.

I'll contest this:

- the experience, as a user, of browsing the page is ~identical under either implementation. For the vast majority of people it is therefore not a "change to the whole system." As an implementation, it is a refactor
- the specific brew is incidental. This move, most importantly, fixes Plutonium (in which footnotes are 100% broken); and (hopefully) alleviates a third-party issue. It does both of these while _also_ simplifying the implementation (at least on the 5etools/Plutonium side) to be inline with various other existing things, which is a win
- if footnotes are being used as something other than the intended "page footnotes" pic'd above, then that is a failure of the brew, not a failure of the system..? There are other tools for similar uses (`@note`, `@homebrew`), and/or, as per above, this is "refactor" not "rework" so everything should go on functionally working as it always has

TheGiddyLimit:
as an aside I think `@footnote` was literally added for Strongholds & Followers, so rejigging it to suit Flee Mortals seems apt 😼

ebullient:
> apologies; smashed out meaningless junk before going to bed. What I meant was: {@sup {@footnote Spell Action Note|FleeMortals|A}} which is essentially what you said but wrapped in a styling tag (currently valid syntax elsewhere, e.g. {@i {@spell ...}})

Frankly, I think this is worse, and it really messes up what are otherwise functional footnotes everywhere else. This is the only source that does things this way, and I feel like you grabbed the footnote as the closest thing to what you wanted, rather than having it make sense.

The system now leverages what most Markdown systems can already do (generating footnote numbers so folks don't have to maintain a count.

> as an aside I think `@footnote` was literally added for Strongholds & Followers, so rejigging it to suit Flee Mortals seems apt 😼

The `{@footnote..}` tag has been in the schema as text for a very long time, and is in much wider use. This particular usage (in FleeMortals) is inconsistent. I think you kind of jammed it into a footnote tag because that was close and got you the hover, but that isn't really what it is. Technically, it's a cross-reference to a notation convention in the source material. You could do something like `{@notation .. }`, which would be more appropriate for this use, and leave all of the other footnotes alone.

> I thought the point of footerEntries on a Spellcasting entry was that this is how wizards and inspired brew sources laid out the notes for spell lists?

It seemed footerEntries was the right place for this, which is why I suggested adding it there.

> Changing the whole footnote system which is used very differently in other parts of the site to enable this specific brew feels strange. 

It really does. Most of those other footnotes are very contextual (none of them match each other). I agree that changing all footnotes to deal with this case (where this case is the outlier) seems strange.

ebullient:
That example is actually a little tough, as it also defines what it thinks the footnote should be. ;)

Here is a more typical one: 

image

Although (in this case), the text is actually the same (this is _also_ in spells, so I feel like this would be another candidate for `{@notation...}`:

image

The source: <https://raw.githubusercontent.com/TheGiddyLimit/homebrew/master/creature/Kobold%20Press%3B%20Creature%20Codex.json>

ebullient:
From this one: https://raw.githubusercontent.com/TheGiddyLimit/homebrew/master/adventure/JVC%20Parry%3B%20Call%20from%20the%20Deep.json

We have text like this: 

> "Within the settlements of the Sword Coast one can find humans, dwarves, elves, and other civilised races commingling more or less peacefully. Waterdeep, in particular, is a melting pot of races from all over Faerûn. Outside of these communities, however, people have much less tolerance for other races, as folk tend to feel safer among their own kind. Small towns and villages dominated by humans tend to have few, if any, non-humans, with most dwarves, elves, and halflings preferring to live in their own settlements, far from human-claimed lands. In generations past, most human settlers of the Island Kingdoms were fair-haired and light-skinned. Since then, the riches and promise of the islands have attracted folk from other regions, and several generations of cultural mingling has given the humans of Faerûn much greater diversity in their appearance (see \"{@footnote {@race human|phb|Human Names and Ethnicities}|See the {@i Info} tab of the {@race human} race.}\" in chapter 2 of the {@i Player's Handbook})."

Which would appear as above. The footnote number would be generated (relative to the note), with footnotes at the bottom in the same way as shown above.

The same is also true for footnotes in `collection/Kobold Press; Deep Magic 17 Mythos Magic.json`, `collection/Kobold Press; Tome of Heroes.json`, as well as `collection/MCDM Productions; Strongholds and Followers.json` (which you noted earlier).

ebullient:
> As for the markdown, all looks reasonable to me. I imagine explicitly differentiating `@notation` and `@footnote` could help you format things nicely? 

Yes!

> That said, as this is a third-party use of """internal""" data this is not _directly_ my concern...

😝

> I think the question then is, as I've been shouted back from the cliff-edge on the `@footnote` rework, do we like the idea of a `@notation` + `"notation": [...]` entity type? It's "free" enough to add/implement on my side, so I've no technical reason against doing so. It has a well-enough-defined use-case/precedence that it makes sense in the bigger picture?

I think the `@notation` type would match to `"notation": [...]` very well, yes. Would allow reusable conventions for the whole source. Solves your Flee Mortals thing nicely, and means the meaning of A, and + is only defined once, and then reused/referenced throughout.

revilowaldow:
Sounds good to me, apologies if I've been shouting, 2°C Monday morning innit.

I won't be satisfied until get a 
`"refstyle": [ Harvard, APA, IEEE, MLA, MHRA ]`

TheGiddyLimit:
coolcool, will see to this after the new book's in and done, so will leave this PR until ~the weekend and close it when I've made the `@notation` (final name pending) version.

> apologies

not at all; please do always shout. I cope by CSSing a funny moustache over your avatar.

> `"refstyle": [ Harvard, APA, IEEE, MLA, MHRA ]`

😩

PR #1885, opened 2023-10-21
Add files via upload

JamesMcGowan123:
After much toil and trouble, the early version of our witch class is finally here and available to you, our beloved patrons!! A lot of care and deliberation went into developing the class for Erika to use, and we’ve been so eager to share it with you all and hear what you think.

We have to stress that this is an early draft of the class mechanics—it hasn’t been thoroughly playtested (you’ve basically heard the extent of it on the podcast), so GMs should be fully aware of that before allowing it to be used in their games.

Because of this, we’ll be seeking feedback from the Patreon in a month or two, once people have had some time to play the class and test it out. We don’t expect feedback, but if you decide to try it out and are willing to share some thoughts, we’d love to hear from you, so keep an eye out!

Big thanks to everyone who helped us put this together, including designers Mazey Veselak, Brandes Stoddard, and Hannah Rose; layout artist Ruby Lavin; illustrators Corey Brickley, Tucker Donovan, and Taylor Moore; and original character illustrator Lorena Lammer.


revilowaldow:
File is in the wrong folder, should be moved to `class/`

p4535992:
I think this can be closed https://github.com/TheGiddyLimit/homebrew/blob/master/class/Worlds%20Beyond%20Number%3B%20Witch%20(Legacy).json ?

PR #1897, opened 2023-11-06
LaserLlama; Savant: Foundry & formatting

hakr14:
not sure why github wants to include 18 effectively empty commits with this but okay i guess


TheGiddyLimit (approved):

TheGiddyLimit:
just to note, the Foundry description HTML gets clobbered by the sanitizer. This isn't something which I'm too keen on changing, but haven't looked into workarounds. Can one use Markdown instead..?

PR #1905, opened 2023-11-14
Kobold Press; Tome of Heroes

Matt-Repinski:
Original converter left off one Eldritch Invocation so I added it in.


TheGiddyLimit (approved):
fine work, but please avoid reformatting the entire file in future

Matt-Repinski:
Apologies. Just used some online tools to make the changes. Hadn't worked with coding at all before.

PR #1906, opened 2023-11-16
Update Middle Finger of Vecna; Alchemist Redux.json

K-S-F:


p4535992:
The issue can be closed https://github.com/TheGiddyLimit/homebrew/blob/master/class/Middle%20Finger%20of%20Vecna%3B%20Alchemist%20Redux.json

TheGiddyLimit:
tytyty

PR #1924, opened 2023-12-04
Aura HB

Aura-6931:
Creatures HB'd by Aura


TheGiddyLimit (changes requested):
broken/truncated filename? Should at least be `.json`, but it looks like there's more missing

PR #1948, opened 2023-12-30
BREW-694: Restore Subclasses to Revised Ranger

revilowaldow:
Adds official subclasses back to the revised ranger UA


TheGiddyLimit (approved):
tyvm

PR #1966, opened 2024-01-20
Update CftD token images

Spappz:
Three creatures in here use the same tokens as WotC creatures due to their statblocks being variations of them. However, because they're _substantial_ variations, they're not `_copy`. Migrating one of these from `mirror-1` and _two_ of these from `noads`!

If there's a better way to point `tokenUrl` to an internal image please let me know 👍

### DIFF
--- a/adventure/JVC Parry; Call from the Deep.json
+++ b/adventure/JVC Parry; Call from the Deep.json
@@ -24732,7 +24732,7 @@
 				"name": "Merrenoloth",
 				"source": "MTF"
 			},
-			"tokenUrl": "https://noads.5e.tools/img/MTF/Merrenoloth.png",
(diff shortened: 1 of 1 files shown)


revilowaldow:
I guess one of the simplest ways to let art referencing work would be to repurpose whatever `altArt` does with name+source type refs however `altArt` currently only works if there's primary art. 

```json
"tokenUrl": "https://i.imgur.com/w9nTza9.png",
"altArt": [
      {
	      "name": "Black King",
	      "source": "TDCSR"
      }
],
```

image

TheGiddyLimit (approved):

TheGiddyLimit:
@Spappz afraid not; if you fancy sticking in a suggestion, go for it

revilowaldow:
I'd be voting for this, I have quite a few brews pointing at mirror/site that would benefit.
There's nearly 400 refs that could be cleaned up in the repo.

image
image

PR #1992, opened 2024-02-04
Added Arcanum Worlds; Odyssey of the Dragonlords

shadowslasher410:
I hope I'm doing this right... Anyway, all of the book except some of the appendices is there.


TheGiddyLimit (changes requested):
couple things:

- please avoid making changes to `package-lock.json` (preferably amend your commit(s) to remove those changes; but reverting them is also fine)
- please don't shorten `json` sources; `ArcanumWorldsOdysseyoftheDragonlords` was an ideal name as-is. If that's too long, then `OdysseyoftheDragonlords` is also fine

TheGiddyLimit (changes requested):
getting there, but the image dir is still named `OotD`; please rename it to match the JSON source also
(and presumably find-replace all the image paths)

TheGiddyLimit (approved):

briavers:
You good sir, Thank you

PR #1993, opened 2024-02-06
BREW-573 Pathfinder Kingmaker Bestiary (5E)

revilowaldow:


revilowaldow:
Slight diversion here for fixing media urls.

Suggest after this is merged adding a `"pattern": "\\.(webp|mp3)$"` to `mediaHrefInternal` in the schema as there are no non-webp/mp3 assets currently referred to by the site or homebrew.

TheGiddyLimit (approved):

PR #1997, opened 2024-02-10
Fix relative image URLs

Spappz:
There's three things occurring here.

## Migrating _relative-path_ `tokenUrl` to `token`
I noticed a bunch of brews used `tokenUrl` as a relative path to reference core images. For instance:
```js
"tokenUrl": "/img/MM/Basilisk.png"
```
This clearly _once_ worked (because I used it 😏) but, naturally, not any more. Since they're all actually-properly-broken-for-real (unlike `tokenUrl`), I felt it was worth migrating them to the new `token`-object format early.

This stuff was pretty simple; check `collection/badooga; Elder Evils.json` for an example.

## Updating stragglers from the grand 5etools `webp`ification
Clearly you forgot relative paths exist too 😏

This is the bulk of the diff. Not much to say; everything relevant is just now webp. There were a few cases of fixing `image` objects that were listed as `href.type = "external"`, but that's all minor.

## Removing dead links
A few brews had relative paths to something that was obviously meant to be a homebrew file. In each case, I couldn't find _anything_ close still in the repo or 5etools-img, so... they're getting deleted. These are the files if you want to double-check:
- `collection/JasonNewby; Durnan's Guide to Tavernkeeping.json`
- `collection/Monte Cook; Arcana of the Ancients.json`
- `collection/QL Games; Tasha's Crucible of Everything Else - Volume 1.json`
- `collection/Yorviing; Yorviing's Arcane Grimoire.json`
- `creature/Onyx Path; Scarred Lands Creature Collection.json`

---

## Thoughts
Might be a good idea to add a guard against `image`s with `href.type = "external"` not targetting an external image? For instance, you could add a regex-match requirement of `^https?://`. Once all internal-facing URLs have migrated, too, you can make the guard more stringent: must now match `^https?://(?!(5e\.tools|5etools-mirror-\d)\b)`.

I also noticed that `audioClip` correctly directs the `path` to the site repo rather than 5etools-img, so nothing needed to be changed there. However, if you wanted to move them at a later date, just be aware that there's a lot of both direct and relative links pointing to them. (You might also want to consider moving direct `external`-link references to those audio clips to an internal-path representation too, if you're already doing it for images.)


TheGiddyLimit (approved):
> forgot relative paths

what kind of idiot--

I see.

Great work, tyvm!

PR #2002, opened 2024-02-12
Add new Haste Knight Fighter subclass

un1plicate:
- New Haste Knight Fighter class.


TheGiddyLimit (changes requested):
please avoid updating the `_generated` files, if you can amend the commit that'd be ideal

additionally, please use a longer `json` source name, e.g. `HasteKnight`

p4535992:
The issue can be closed https://github.com/TheGiddyLimit/homebrew/blob/master/subclass/Jagans_eye%3B%20Haste%20Knight.json

PR #2007, opened 2024-02-18
Adding subclasses

shirayuri1:
Adding revised subclasses for a homebrew west march server


TheGiddyLimit (changes requested):
please use longer `json`/`source` names, e.g. `"TierrasDevastadasArcaneMarksman"`

PR #2014, opened 2024-02-20
Status change Wip > Ready

Foopsie42:
Forgot to change the status of the file from "wip" to "ready", so i did that and added the date.


TheGiddyLimit (approved):
that's an awfully large diff for 3 lines of change

Foopsie42:
> that's an awfully large diff for 3 lines of change

I usually don't edit directly in git, editing file in VSC and then replacing the old one in repo. Maybe that's why

PR #2022, opened 2024-02-28
Remove UA Fighting Styles

badooga:
Removes references to UA Fighting Styles from two of my brews, which would cause the classes to entirely just not load


TheGiddyLimit (approved):
my bad, thought I cleaned those up

PR #2032, opened 2024-03-03
add subclasses

shirayuri1:
These are reworked subclasses for a westmarch campaign
Assassin
Berserker
Champion
Thief
Arcane Archer


TheGiddyLimit (changes requested):
looks good, but the files would be better named

`subclass/Shirayuri; Tierras Devastadas Arcane Dead Shot.json`

etc.; the "full" source usually goes in the filename rather than the "json" source

TheGiddyLimit (approved):

PR #2049, opened 2024-03-15
FIX AGAIN

revilowaldow:
This is yet more fixes for Odyssey of the Dragonlords.

@shadowslasher410 please can you check as you go, and ideally use this file as the starting point for any further work you do


shadowslasher410:
Will check out in depth tomorrow (have D&D session in half an hour. Regardless of the JSON, there are a couple of images with white borders that I fixed that I need to commit as well, which I will when I finish checking the new one out).

revilowaldow:
No worries. FYI the first commit just automatically puts everything in a standard sorted order so it's easier to find things. The second commit is the actual fixes 👍 enjoy your session

shadowslasher410:
Wait... there's a standard sorted order? I just did it alphabetically... Where would I find that? Is it in the schema?

revilowaldow:
It's a feature of the site that requires you have a copy of that downloaded as well. It's a bit more nuanced than "just alphabetical", for example class features are sorted by:
- class source
- then class name
- then feature level
- then feature name
- then feature source
 so they're in a sensible order.

image

revilowaldow:
https://github.com/5etools-mirror-2/5etools-mirror-2.github.io/blob/main/js/utils-proporder.js

shadowslasher410:
So, I fixed/added the changes and updated my fork, but for some stupid reason, it has some other files changed too even though I didn't touch them. The only ones that should be changed are the image files and the OOTD JSON.

Oh, and as an aside, I'm pretty sure the homebrew's package.json needs to be updated too, with the latest versions of commander and 5eutils.

TheGiddyLimit (approved):

TheGiddyLimit:
my bad; didn't bother to check out the mega-diff locally, and it doesn't load in a browser

PR #2087, opened 2024-04-17
Add files via upload

Arkroup:


hakr14:
Homebrew files are not allowed in root. I'd recommend moving the file to the `collection` directory.

TheGiddyLimit:
Merged in #2375

PR #2112, opened 2024-05-06
Update Dusk Zargoth; Monk Rework.json

DuskZargoth:
I had to update some minor details, but there's still work to be done. Currently dealing with what appears to be a bug in the system


DuskZargoth:
Current changes ready

TheGiddyLimit:
not quite; you'll need to resolve those merge conflicts!

DuskZargoth:
I don't understand these merges conflicts.

DuskZargoth:
Like, why are they in conflict in the first place? I have tested the raw link in Foundry and it didn't have an issue. And regarding the indentation, it's a problem with Github, because in the dev tool it looks correct.
Captura

DuskZargoth:
Why are you asking me to replace the subclass description for another from a totally different one?
Also, the <foundryImage> for the subclass HAS to go in the <subclass>, it doesn't work if I put it in the <subclassFeature>
imagen
imagen

DuskZargoth:
I still corrected the indentation as suggested, and only the indentation, the rest shouldn't be considered an issue because that's how the schema works. Hold on, I think I deleted the goddamn codes for the spell of my Shadows subclass.

PR #2121, opened 2024-05-14
Update Monk Rework and Add Fighter Class

DuskZargoth:
I'm not sure if this is the correct process, I have tried multiple ones. So I'm gonna stick with this one.


TheGiddyLimit (approved):

TheGiddyLimit:
if it works it works

PR #2139, opened 2024-06-11
Adding The Flying Misfortune by Eventy Games

willioz:


hakr14:
Your json source is invalid. It must be an alphanumeric string (with special characters [- :&+!]) at least 6 characters long. [Relevant part of schema](https://github.com/TheGiddyLimit/5etools-utils/blob/a3df5e07256d961d6955f087ba56a070c35c4a5b/schema/brew/util.json#L48C3-L63C4)

PR #2144, opened 2024-06-15
Create CaelReader; All the Weapons.json

hakr14:
i don't care anymore


TheGiddyLimit (changes requested):
images need to be placed in a dir which matches the source `json`; have moved your `AtW` to https://github.com/TheGiddyLimit/homebrew-img/tree/main/img/atweapons if you want to update this PR with appropriate paths

PR #2145, opened 2024-06-15
Add files via upload

Masterplaster042:


TheGiddyLimit (changes requested):
`Error: File "class" array is empty!`

looks like you need to neaten up your file (note that there may be _other_ empty arrays which it'll also complain about if you don't fix, too!)

PR #2147, opened 2024-06-21
Chronicle of Heroes; The Weaveknight (v4.0) + Complete Arcane Player's Guide

Spappz:
For fantom 😊

Interestingly, two files already existed for *The Weaveknight*, one for v2.5 (complete) and one for v4 (incomplete). I finished off the latter and marked the former as `deprecated`; I've also marked the v4 version as `unlisted` to match v2.5 (you can search for logs on the topic in Discord).


TheGiddyLimit (approved):
I will instead simply trust you know what you're on about

(sounds good)

PR #2148, opened 2024-06-21
Y's Codex of Realities

RainerRaccoon:
Embrace the changes and additions that brought about new gods, saved the world multiple times, and made some of the best memories this group has ever had. Read about and create characters and stories with the content surrounding the unforgettable tales all packaged within the JCU. All that and more are documented for later use within this book. Whether the JCU is a world you are actively interacting with or one you'd like to passively read about, this book is sure to give the mind plenty to image and create with.


TheGiddyLimit (changes requested):
tests are (rightly) complaining that this does not include any book content; you'll want to move it to `collection/` instead of `book/`

PR #2149, opened 2024-06-22
Conflux Creatures; Conflux Creatures

Spappz:
[Conflux Creatures](https://www.patreon.com/confluxcreatures/) has kindly made all their stuff free for the weekend, and has _also_ included a 5eT homebrew file as part of that. We apparently already had an older version here, so I've gone and updated it and done... quite a bit of cleaning (by no means near complete; I've just called it a day since this is stuff the guy usually charges for).

I've left my name off the `convertedBy` list since, well, I didn't really do the conversion, and, more importantly, I don't want to be bugged about it 😏


TheGiddyLimit (approved):
based

PR #2156, opened 2024-07-01
Wraith Wright's Comprehensive Equipment Manual

Billiam9420:


p4535992:
Is a book right ? it should be put under the "book" folder not the "item" folder

Billiam9420:
> Is a book right ? it should be put under the "book" folder not the "item" folder

It's a book focused on items, so I felt item was appropriate.

TheGiddyLimit:
given that ~40% of the file _is_ `"bookData"`, it probably does fit better under `book/`. Since there's no clear guidance/enforcement, though, this is fine for now

TheGiddyLimit (approved):

PR #2209, opened 2024-08-07
Kobold Press ToH fixes; Deep Magic fixes, 2nd printing updates, and page numbers

TurqW:
Just fixing various stuff that came up while I was deep diving some builds.

1. Two spells had their range, materials, etc swapped
2. Added page #'s for all Deep Magic subclasses & subclass features so their source is obvious
3. I don't have a first printing copy, just a brand new PDF that appears to be based on the second printing, but a few features seem to have been renamed and tweaked. Where I noticed, I switched them over to match the newest edition.
4. Greenleaf Bard previously displayed one feature twice
5. Random extra text from another subclass in the Aristocratic Sorc description
6. One spell had a cyrilic character as its school instead of a latin C that would be correctly translated to Conjuration

I've probably missed something in this enumeration, sorry for mixing real fixes in with the page #'s, it didn't occur to me until just now that I probably should have split the commits.


TheGiddyLimit (changes requested):
looks like you've got a junk `\n` character somewhere

TurqW:
Fixed. Thanks, PyCharm 🙄

TheGiddyLimit (approved):

PR #2210, opened 2024-08-07
Add images for all subclasses in Griffon's Saddlebag Book 1

unknown117343:
While normally I would upload images to the homebrew image repo, this homebrew appears to be legacy in that it's images aren't in the homebrew image repo but rather in this repository. For the sake of keeping all the images pertaining to this homebrew together, I have opted to upload the subclass images in this repo instead.


p4535992:
Just as knowledge for images related to home brew resources there is a dedicated git project here: https://github.com/TheGiddyLimit/homebrew-img is not mandatory only advisable

PR #2211, opened 2024-08-10
Griffon's Saddlebag Book 3

unknown117343:
Added subclasses & consolidated Year 4 & 5 items in Book 3. All of which have images.

I still have to add monster stat blocks, settings, and two races but that will come later.


unknown117343:
Unclear to me why this is failing. Live schema testing & test script both passed. Looks like the failure resides with another JSON since the error mentions a monster but I don't have any monsters in this yet.

TheGiddyLimit:
yeah my bad, made a schema change without double-checking the results

TheGiddyLimit (approved):

PR #2212, opened 2024-08-10
Delete item/griff-mac; The Griffon's Saddlebag.json

unknown117343:
This no longer needs to exist. There are a small handful of items from the Griffon's Saddlebag included in here. All of which are covered in Book 1.


revilowaldow:
Rather than deleting something in this case we prefer to give it the `"status": "deprecated"` tag in the `_meta`.
This hides it from the homebrew options by default but is still there should it need to be referred to in future, or someone just wants the old version.
image

TheGiddyLimit:
will deprecate instead

PR #2219, opened 2024-08-17
Add SCGtD subclasses and contaminated spells

dkparker:
Hey, I converted the subclasses and contaminated spells from Sebastian Crowe's Guide to Drakkenheim. This is my first time doing this, so I'd appreciate the review of someone who actually knows what they're doing. 

Everything passes the schema check afaik. 

Not sure why the diff is all messed up, but I just added everything new after what was already there. 

I tested by using file import on foundry and importing class features and spells to characters. 

One issue I had was that sometimes the class feature references wouldn't find the feature unless I changed the name of the class feature to something else, which is why the parent class features sometimes have weird or abbreviated names.

Also, trying to import the delerium elemental or the thing with the writhing tail just hung forever when I tried in foundry, not sure why that was happening.

Finally, I wasn't able to get everything fully automated with my limited knowledge.

Since there are those and perhaps other unknown issues, this is a WIP PR and I'm of course happy to make any edits as appropriate, as long as I can get help figuring out how to make those edits. Alternatively, if you want to just take this first draft and do your own edits, that's fine by me too.


TheGiddyLimit:
> One issue I had was that sometimes the class feature references wouldn't find the feature unless I changed the name of the class feature to something else, which is why the parent class features sometimes have weird or abbreviated names.

that definitely shouldn't happen. Can you update this to use the original names, and I'll take a look?

fwiw if I try to import anything from the file currently, it crashes

dkparker:
Not sure what to tell you, I switched the parent subclass features back to the normal name and it worked. No clue what could have caused that. 

I realized the commit I pushed had a genuine broken reference though, which I fixed. Sorry, not sure how that slipped in.

collection/Dungeon Dudes; Sebastian Crowe's Guide to Drakkenheim.json
```diff
@@ -2865,248 +3440,3632 @@
...
+				"activation": {
+					"type": "bonus",
+					"cost": 1
+				},
+				"save": {
+					"ability": "cha",
+					"scaling": "cha"
+				},
+				"uses": {
+					"max": "floor(@prof / 2)",
```
dkparker, line 3866: For some reason, when imported, this still gives 2.5 uses instead of 2 when PB is 5. Any idea why that would be?

dkparker:
Thank you for all your hard work. Plutonium and specifically the homebrew collection have made the games I run so much better.

PR #2221, opened 2024-08-21
da-mage

revilowaldow:


TheGiddyLimit (changes requested):
should be `\\b` rather than `\b` (I think), as otherwise it's an escaped `b` character

I usually do both sides too, to complete the ugly: `\\bmage\\b`

revilowaldow:
Aaaaaaaaa 

Fair, I think sometimes you'd want `mage's` to match so the second \\b may be unideal as a general rule. Though yes double escape likely important

TheGiddyLimit:
re:

> want `mage's` to match

image

works on my machine 👀

revilowaldow:
Clearly need to brush up on my regexp knowledge

PR #2224, opened 2024-08-23
Added Griffons Saddlebag 2 subclass & race images

unknown117343:
@mfried9 Heads up

### DIFF
--- a/collection/Griffin Macaulay; The Griffon's Saddlebag, Book 2.json
+++ b/collection/Griffin Macaulay; The Griffon's Saddlebag, Book 2.json
@@ -21658,6 +21658,15 @@
 							}
 						]
 					}
+				],
+				"images": [
+					{
+						"type": "image",
+						"href": {
+							"type": "external",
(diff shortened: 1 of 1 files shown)


collection/Griffin Macaulay; The Griffon's Saddlebag, Book 2.json
```diff
@@ -22107,6 +22125,200 @@
 			]
 		}
 	],
+	"subclassFluff": [
+		{
+			"name": "Path of the Infernal",
```
revilowaldow, line 22130: None of these will be picked up without accompanying `"hasFluffImages": true` on each `subclass`

revilowaldow (changes requested):

collection/Griffin Macaulay; The Griffon's Saddlebag, Book 2.json
unknown117343: Picked up by what? Plutonium? Looks fine in 5etools. (Regardless, I think I am going to move everything from subclassfluff as fluff under subclass).
image

PR #2238, opened 2024-09-03
Blood Hunter subclass

Ark1002:
Order of the Revenant for the Blood Hunter Class, plus a blood curse.

### DIFF
--- a/collection/Order of the Revenant.json
+++ b/collection/Order of the Revenant.json
@@ -1,0 +1,185 @@
+{
+	"_meta": {
+		"sources": [
+			{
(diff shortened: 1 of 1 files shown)


TheGiddyLimit:
filename needs to be `Ark1002; Order of the Revenant.json`, if you're the `author` 

---

you might also want to stick a:

```
"includes": {
	"class": [
		"BH2022"
	]
},
```

in the meta to ensure the base class gets loaded

TheGiddyLimit (changes requested):
.

PR #2248, opened 2024-09-14
Oath of sanctuary

Haveltheblock:


subclass/Arbiter; Oath of Sanctuary .json
```diff
@@ -0,0 +1,264 @@
...
+							"entries": [
+								"You must do everything possible to protect those in your care, even if it means sacrificing yourself or others outside your care. Putting those under your care comes above you at all costs."
+							]
+						},
+						{
+							"name": "Loyalty",
+							"entries": [
+								"Once you have committed yourself to protecting an individual, you must protect them until the terms of your contract or other agreement are done. You cannot willingly end it."
+							]
+						}
```
hakr14, line 126: The subentries in Tenets of Sanctuary are missing types:
```suggestion
						{
							"name": "Protection",
							"type": "entries",
							"entries": [
								"Your first and foremost responsibility is to protect everyone in your immediate care. You must remain at the side of those entrusted to your care rather than seek revenge or justice."
							]
						},
						{
							"name": "At All Costs",
							"type": "entries",
							"entries": [
								"You must do everything possible to protect those in your care, even if it means sacrificing yourself or others outside your care. Putting those under your care comes above you at all costs."
							]
						},
						{
							"name": "Loyalty",
							"type": "entries",
							"entries": [
								"Once you have committed yourself to protecting an individual, you must protect them until the terms of your contract or other agreement are done. You cannot willingly end it."
							]
						}
```

hakr14 (changes requested):

subclass/Arbiter; Oath of Sanctuary .json
Haveltheblock: Oh whoops

subclass/Arbiter; Oath of Sanctuary .json
```diff
@@ -0,0 +1,267 @@
+{
+	"_meta": {
+		"sources": [
+			{
+				"json": "Havels Random Finds",
```
hakr14, line 5: This json identifier is already in use: https://github.com/TheGiddyLimit/homebrew/blob/fc84c6a5e77c9b96cd694db30989bd6a7636ae00/subclass/Piggby%3B%20Oath%20of%20Ashes.json#L5

Haveltheblock: Oh I gotta change that to? Even though thats still me?

hakr14: Yes, multiple sources can't have the same `json`. You'd need to add this to the same source, and you can only do that in the same file.

Haveltheblock: I just changed it. Turned green though so im good right?

TheGiddyLimit (approved):

PR #2256, opened 2024-09-19
Cleanup the Deep Magic collection from Kobold Press

p4535992:
- Renamed source codes to at least 6 digits as recommended
- Unified the version mechanism with @ YYYY
- Updated the urls to point to the official kobold press site
- Added Deep Magic 18 (finalized the one started by Spappz)

I don't know if I made the right choices, I put the PR in draft pending confirmation.


revilowaldow:
- I wouldn't change established json identifiers as it breaks things like the content updater, duplicate handling etc in plutonium.
- If any of those images are new they should go in the dedicated homebrew assets repo
- If they're the same, leave them where they are to avoid breaking people's games in roll20/foundry where assets would suddenly be missing
- No comment on the addition, presume that's fine

p4535992:
- Revert images move
- Revert the renamed source codes (with the exception of “dm17” which I capitalized all “DM17” like the others, but it shouldn't cause any problems)
- Unified the version mechanism with @ YYYY
- Updated the urls to point to the official kobold press site
- Added Deep Magic 18 (finalized the one started by Spappz)

revilowaldow:
It's failing on DM17 having been capitalised as the legacy source ID is not, and it's shorter than 6 characters. Understand the case change shouldn't matter much, but that's why it's failed.

p4535992:
I'll check out the last error tonight, sorry for the toruble.

p4535992:
- Unified the version mechanism with @ YYYY
- Updated the urls to point to the official kobold press site
- Added Deep Magic 18 (finalized the one started by Spappz)

TheGiddyLimit:
(apologies for the main-branch merge; jogging the test runner)

TheGiddyLimit (approved):

PR #2267, opened 2024-10-03
Add new homebrew

FlamingQaz:
Converted Blue4Eternety's Fallen Undead Race Homebrew

### DIFF
--- a/race/Blue4Eternety; Fallen - A True Awakened Undead Race .json
+++ b/race/Blue4Eternety; Fallen - A True Awakened Undead Race .json
@@ -1,0 +1,462 @@
+{
+	"_meta": {
+		"sources": [
+			{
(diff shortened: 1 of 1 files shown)


race/Blue4Eternety; Fallen - A True Awakened Undead Race .json
```diff
@@ -0,0 +1,462 @@
+{
+	"_meta": {
+		"sources": [
+			{
+				"json": "FTAUR:Fallen",
```
TheGiddyLimit, line 5: afraid `:` isn't allowed in `json`/`source`s; please replace with e.g. `FTAURFallen` throughout

TheGiddyLimit (changes requested):

PR #2269, opened 2024-10-05
Overhaul Arcadia 2

revilowaldow:
Overhaul Arcadia 2

### DIFF
--- a/book/MCDM Productions; Arcadia Issue 2.json
+++ b/book/MCDM Productions; Arcadia Issue 2.json
@@ -21,6 +21,508 @@
 		"dateLastModified": 1705266948,
 		"_dateLastModifiedHash": "23b138019f"
 	},
+	"subclass": [
+		{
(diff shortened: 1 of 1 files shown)


TheGiddyLimit (approved):

TheGiddyLimit:
for these arcadia updates, how many of the images are new? 

cos In retrospect, it would've been good to long-ify the source names to e.g. `Arcadia<X>`, and take them off the "reserved" list 😩 Mb for not calling it out sooner

revilowaldow:
Re images it depends quite a lot. The arcadias were converted by like 6/7 different people so they're all different quality levels. Lots of token additions, some missing images, some existing stuff getting removed as they're just screenshots of whole pages.

revilowaldow:
Currently Ar1-8 and Ar20 are in the legacy list so there would be a lot of source changes. Though I think about half of the 30 total are converted.

PR #2277, opened 2024-10-28
Create SilentSoren; Endwalker Compendium - Final Version

SilverStrings5526:
Silent Soren's FF compendium, backloaded to endwalker because my players are still using it and the new version breaks their sheets


TheGiddyLimit (changes requested):
using the same source (`FFXIV`) as the other file won't work, you'll need to swap that out

there's also a length requirement now, so whatever you change it to needs to be >= 6 characters long

I'd recommend e.g. `FFXIVBeforeTheFall `

SilverStrings5526:
would i need to replace every instance of "FFXIV" or only when its preceded by "source"
image
image

TheGiddyLimit:
every instance, I'm afraid

TheGiddyLimit (changes requested):
> `File did not have ".json" extension: SilentSoren; Endwalker Compendium - Final Version`

simple enough; file needs renamed to have `.json` on the end

TheGiddyLimit (changes requested):
`"json": "FFXIVBeforetheFallBeforetheFall",`

looks like you've double-find-replace'd the `json` part on line 5, and it's breaking things

TheGiddyLimit:
needs e.g. `"edition": "classic"` in the `_meta`; see the original file: https://github.com/TheGiddyLimit/homebrew/blob/master/collection/SilentSoren%3B%205th%20Edition%20D%26D%20x%20Final%20Fantasy%20XIV%20-%20Classes%20and%20Races%20Compendium.json#L38

TheGiddyLimit (changes requested):
invalid JSON now eh

https://jsonlint.com/

TheGiddyLimit (changes requested):
as much as I enjoy being the go-between between the tests and this PR, you might consider running `npm t` locally until it passes

instructions should be available on the [wiki](https://wiki.tercept.net/en/Homebrew/FromZeroToHero), see e.g. "Testing your JSON"

SilverStrings5526:
hope this one works cause VSC shows all clear with the language server plugin

TheGiddyLimit (approved):

PR #2295, opened 2024-11-24
Selerim

jknolte:


TheGiddyLimit (changes requested):
your `hasFluff`/`hasFluffImages` stuff should be in the background; currently it's in the root of the file

see e.g. https://wiki.tercept.net/en/Homebrew/FromZeroToHero for how to check locally

jknolte:
That's helpful. I realized I don't need to actually load my homebrew to the repository in order to use it on my Foundry Server, so I'm going to abandon this for now.

PR #2316, opened 2024-12-11
Custom Backgrounds

DaveParkinglot:
These backgrounds are basically the 2024 backgrounds but I added suggested characteristics similar to the 2014 backgrounds :)

Also sorry for all the pulls, I'm very bad at Github


TheGiddyLimit (approved):

TheGiddyLimit:
would still be better with a bit of `_copy` abuse, but it'll do

DaveParkinglot:
> would still be better with a bit of `_copy` abuse, but it'll do

I figured having the Source be different would be enough. Plus if I had "Acolyte_Copy" in my game, my players would get confused lol

TheGiddyLimit:
`_copy` as in the thing that allows you to copy and modify an existing entity:

- https://github.com/5etools-mirror-3/5etools-src/blob/main/data/backgrounds.json#L2278
- https://github.com/TheGiddyLimit/homebrew/blob/master/_doc/Spec_Copy.md

DaveParkinglot:
Oh good call, I didn't realize it could be that easy. I'll take a look

PR #2318, opened 2024-12-17
Update Poison Potion Press; Shore of Dreams.json

abbaccus:
Description for tempest beast was used for the description of the storm mephit.  I updated it with the basic mephit description since there is no description information about storm mephits in the source adventure.

### DIFF
--- a/adventure/Poison Potion Press; Shore of Dreams.json
+++ b/adventure/Poison Potion Press; Shore of Dreams.json
@@ -1533,7 +1533,43 @@
 			],
 			"fluff": {
 				"entries": [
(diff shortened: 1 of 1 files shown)


TheGiddyLimit (changes requested):
better would be to remove the `fluff` from the Storm Mephit, and add a copy of the generic Mephit fluff like so:

```jsonc
// Merge with the existing `_meta`
"_meta": {
	"dependencies": {
		"monsterFluff": [
			"MM"
		]
	}
},

// Add to the file's root
"monsterFluff": [
	{
		"name": "Storm Mephit",
		"source": "SOD",
		"_copy": {
			"name": "Mephits",
			"source": "MM"
		}
	}
]
```

PR #2324, opened 2024-12-24
Updated Storm Mephit

abbaccus:
Updated Storm Mephit to use the Mephits monsterFluff from MM, instead of erroneous description of tempest beast.


abbaccus:
Thanks for the direction on how to reference monsterFluff.  I am sorry for the extra commits; I am new to git.  I was testing locally to make sure things worked before submitting the pull request, and I was seeing some issues I was trying to resolve.  Turns out things were just out of sync.  I think I have a better strategy moving forward that will prevent the extra commits.

TheGiddyLimit (approved):
🎉

PR #2404, opened 2025-03-15
Update Dragonix; Monster Manual Expanded II (v2).json

Spappz:


TheGiddyLimit (approved):
fwiw the `.` at the end of `item` `name`s is auto-added by the renderer, should you fancy omitting it

PR #2419, opened 2025-03-30
2024 Custom Background by yours truly

alterNERDtive:
See suggestion `PLUT-527` which is probably not very high priority. This can be used in the meantime.


TheGiddyLimit (approved):
👍 it'll certainly work

PR #2427, opened 2025-04-17
Add Gunner class from "Zaman's Guide to the End of Time" by Loot Tavern Publishing (BREW-822)

Alexandre-Pe:
I don't know why old commits are still shown even though they were included in previous pull request...


TheGiddyLimit (approved):
commits get squashed on merge, so _technically_ the PR becomes its own, unique, commit

the best way to avoid the spam is to branch off a the latest `master` after your previous work got merged

PR #2430, opened 2025-04-23
Update Dungeon Dad; Dungeon Dad Statblocks.json

Genocide-Games:
updated some things here as always, even added a new item in the lich skull


TheGiddyLimit (changes requested):
`"basicRules"` is no longer valid in homebrew, so you'll need to remove those lines

PR #2435, opened 2025-04-25
Finally complete Trapper subclass for Ranger, from Heliana's Guide to Monster Hunting (BREW-372)

Alexandre-Pe:
I added the traps as optional features as the trap/hazard type didn't seem appropriate to use prerequisite and stuff, hopefully that was the right choice.


TheGiddyLimit (approved):
seems fine to me

PR #2438, opened 2025-04-26
Migrate `attachedSpells` 

revilowaldow:
Notes:

- There does not appear to be support for:
	- `proporder` support within `attachedSpells` objects
	- Spells that are prepared when attuned, or known for wizards via spellbook items
	- Spells that consume resources to cast, e.g. requires attunement to by a monk and uses ki
	- Spells that can be cast '_once per n time period'_ , e.g. once every 7 days
	- Spells that specify a spellcasting `ability`; wis, cha etc
- ~It is unclear if `#c` / `#n`-level have any affect within `attachedSpells`~
- Where Evil Lives and Arcadia 10 received minor changes similar to other files but were prettified so had large diffs.


TheGiddyLimit:
since this is only for Foundry use at the moment, the idea is that anything more complex gets whacked in `foundryActivities`

essentially, it needs to fit within the existing `additionalSpells` structure without too much fuss, as I don't want a second one of those

`ability` is deffo low-hanging fruit, so will get that in for next release; `resource` maybe too, even if the data doesn't end up getting used

TheGiddyLimit (approved):

PR #2439, opened 2025-04-27
update of dancer class

Dan69UwU:
update of the dancer class from version 1.3 to 1.6 , added some of the base class feature working directly in foundry


TheGiddyLimit (changes requested):
your effects have lots of `disabled: false`, which the schema doesn't allow (as it's implicit)

you might want to run your Foundry data through the tool in the Automation Addon repo: https://github.com/TheGiddyLimit/plutonium-addon-automation/blob/master/CONTRIBUTING.md#bulk-conversion-from-foundry (see the "Bulk conversion from Foundry" header) which handles scrubbing up and slimming down stuff exported from Foundry

PR #2442, opened 2025-05-01
Updates Alternate Barbarian

CrazierIvan:
DNM


CrazierIvan:
Okay, I think we're finally done here.

PR #2443, opened 2025-05-03
dancer class update

Dan69UwU:


TheGiddyLimit (approved):
the Foundry stuff is [still pretty blegh](https://github.com/TheGiddyLimit/homebrew/pull/2439#pullrequestreview-2797744408)

but if it works it works

PR #2455, opened 2025-05-22
Grant 'Bite' to Vulpin as starting equipment

smiller171:


TheGiddyLimit:
rejecting re: discussion in DMs, in favor of a later PR with an activity instead

PR #2492, opened 2025-07-07
Add Dragon Rider class from The Game Master's Book of Legendary Dragons

AntarEspadas:


unknown117343:
This homebrew was last updated in 2022 which makes it a 2014 homebrew but throughout this conversion I see XPHB and the edition is "one" instead of "classic"

PR #2520, opened 2025-08-06
Update KibblesTasty; Inventor.json

nitrodynamite18:
- Fix Inventor class being called "KibblesTasty's Inventor"

- Change the identifier and source name to fit with the other KT classes in the repo

- Fix spells not being added to the Artificer spell list

- Fix some spells incorrectly referencing the Artificer class in their description

### DIFF
--- a/class/KibblesTasty; Inventor.json
+++ b/class/KibblesTasty; Inventor.json
@@ -3,18 +3,19 @@
 	"_meta": {
 		"sources": [
 			{
-				"json": "KibbleInventor",
+				"json": "KT:I",
 				"abbreviation": "KT:I",
-				"full": "KibbleTasty's Inventor",
+				"full": "KibblesTasty: Inventor",
 				"authors": [
-					"/u/KibblesTasty"
+					"KibblesTasty"
 				],
 				"convertedBy": [
 					"Giddy",
 					"Ikaguia",
 					"MBlackX",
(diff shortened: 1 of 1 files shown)


revilowaldow:
> I don't understand why the checks are failing. I've uploaded the file manually to the homebrew manager and everything works perfectly.

You've changed the source identifier from `KibblesInventor` to `kti` but new source IDs shorter than 6 characters are not permitted to avoid conflicts with the site. You can still make the abbreviation say whatever you want but unique JSON identifiers below 6 chars are now reserved for site use. I suggest you revert that change as it will not be accepted in its current form.

nitrodynamite18:
> > I don't understand why the checks are failing. I've uploaded the file manually to the homebrew manager and everything works perfectly.
> 
> You've changed the source identifier from `KibbleInventor` to `kti` but new source IDs shorter than 6 characters are not permitted to avoid conflicts with the site. You can still make the abbreviation say whatever you want but unique JSON identifiers below 6 chars are now reserved for site use. I suggest you revert that change as it will not be accepted in its current form.
> 
> More generally I would advise that changing an established source ID should be avoided as this will break VTT users characters when they try to level them up and Plutonium cannot find the class/subclass info by Name/source pair anymore.

I see. Slightly annoying that we can't make that fit with the other Kibbles content, but I'll revert that change. Thanks for the advice!

PR #2577, opened 2025-10-19
Update “The Ruffian, Revised” by yours truly to v0.2

alterNERDtive:


alterNERDtive:
Wait, you are not allowed to use HTML, like, at all? That’s quite annoying.

PR #2717, opened 2026-04-07
Big Laserllama update

hakr14:
Reworked Exploits Foundry implementation, cleanup Fighter and Warlord, move Feats to separate Compendium, split Exploits and Fighting Styles to separate sources, update Spells


hakr14:
The errors are, to my knowledge, all `forward` activities, which *should* be fine as of 2026/3/1:
https://discord.com/channels/363680385336606740/1477189590577057984

CrazierIvan:
How did you pull it off? Where is flags.ll.exploit.faces defined?

hakr14:
> How did you pull it off? Where is flags.ll.exploit.faces defined?

The `flags` structure is present in (I think?) all documents in Foundry, as a sort of "free-for-all" dumping grounds for stats and values that don't have a dedicated place elsewhere in the architecture. The D&D5e system even uses it itself for some special traits (Melee Weapon Extra Critical Dice, for example, is at `flags.dnd5e.meleeCriticalDamageDice`.)

You can define a flag yourself in an effect, done here in the "Fighter/Rogue/Warlord/etc. Exploits" effect granted by the Exploits feature by the class. Flags can't reference each other sadly, so for Warlord in particular I had to move some assignments around to be a little less intuitive, but oh well.

It's still far from a perfect solution (multiclassing Exploit classes uses some very... particular rules that I can't find a way to automate), but I think it's better than nothing, at least.

TheGiddyLimit (approved):

PR #2722, opened 2026-04-08
Background Fixes

nitrodynamite18:
Part 1 of the good changes from #2711:
- Fix some skill and language proficiencies being capitalized, causing them to be placed first in the filter function instead of alphabetically.
- Fix some PHB tool proficiencies having the source tag, causing duplicate entries in the filter function.
- Fix some tags using abbreviations rather than source JSON properties.
- ~~Fix inconsistent source tag capitalization. Not sure if this has any benefit other than consistency.~~ Unnecessary.
- Fix file names.


collection/1985 Games; Obojima - Tales from the Tall Grass.json
```diff
@@ -7928,10 +7928,10 @@
 			"startingEquipment": [
 				{
 					"_": [
-						"explorer's pack|phb",
-						"chest|phb",
-						"grappling hook|phb",
-						"fishing tackle|phb",
+						"explorer's pack|PHB",
```
TheGiddyLimit: raw UIDs like this are all-lowercase by convention, fwiw, so this is "worse" (and/or may be clobbered back to all-lowercase in future, so I wouldn't worry about changing them)

tbh sorry to say 80% of your changes to this file are casing things in the opposite direction to convention; things in `@tags` _are_ conventionally mixed-cased, but the rest seems to be Not Good

I'll write up the intended conventions in the site's `CONTRIBUTING.md` for next release

TheGiddyLimit (changes requested):
please revert/avoid changing `_generated` files; these are update on "release"

collection/1985 Games; Obojima - Tales from the Tall Grass.json
nitrodynamite18: I see. I'll revert those changes real quick, and I'll make sure future commits don't have that issue. Having a `CONTRIBUTING.md` with the site's conventions will be a massive help.

nitrodynamite18:
All my changes should now be consistent with convention.

TheGiddyLimit (approved):

PR #2724, opened 2026-04-09
Eberron Tune-Up

nitrodynamite18:
Part 2 of the good changes from #2711:
- Fix a number of subclass spell lists that were partially or fully missing.
- Add some data to subclass features.
- Add and fix tags.
- Fix some typos.
- ~~Fix inconsistent source tag capitalization. Not sure if this has any benefit other than consistency.~~ Unnecessary, discussed in #2722.


subclass/Stuart Broz; Wizards of Eberron - Five Arcane Traditions.json
```diff
@@ -80,317 +140,565 @@
...
 			"classSource": "PHB",
 			"page": 9,
+			"additionalSpells": [
+				{
+					"known": {
+						"2": [
+							"sanctuary"
+						]
+					},
+					"expanded": {
```
nitrodynamite18, line 150: I don't know if the subclass is intended to have access to the spells here that aren't already wizard spells, but they are listed in the Stay Aloft feature. I have added them as expanded options just in case.

TheGiddyLimit (approved):

PR #2732, opened 2026-05-07
Mimic Race

HambiWalnut:


collection/HambiWalnut; Mimic Race.json
```diff
@@ -0,0 +1,650 @@
+{
+	"_meta": {
```
TheGiddyLimit, line 2: needs an `"edition": "classic|one"`

TheGiddyLimit (changes requested):

revilowaldow:
Current issue is that the json id/source is less than 6 characters which is the minimum for new homebrews. Suggest `MimicRace` instead of `Mimic`, you can leave the `abbreviation` as `MIMIC`.

revilowaldow:
`"edition": "classic|one",` needs to just be either `"edition": "classic",` or `"edition": "one",` your choice depending on which version of the game this for

HambiWalnut:
Thank you for helping me fix it!

revilowaldow:
> Thank you for helping me fix it!

You're welcome :)

TheGiddyLimit (approved):

PR #2740, opened 2026-05-12
ASoS Tune-Up

nitrodynamite18:


TheGiddyLimit:
🤔 what's the idea behind removing `webp`s and referencing external-served images (i.e., ones which we don't control/can't "guarantee" availability of)?

TheGiddyLimit (changes requested):
(see prior comment about image removal -- ideally these should be migrated to `homebrew-img`, if anything)

TheGiddyLimit (approved):

PR #2741, opened 2026-05-15
Sterling Vermin Tune-Up

nitrodynamite18:
Mostly data cleanup and typo fixes.


TheGiddyLimit (changes requested):
if we're renaming files, prefer renaming them to the name on the cover (e.g. The Magus, The Pugilist)

class/Benjamin Huffman; The Pugilist - 2014.json
```diff
@@ -21,7 +21,7 @@
 			},
 			{
 				"json": "SterlingVermin:Patreon",
-				"abbreviation": "SVMP",
+				"abbreviation": "BH:P",
```
TheGiddyLimit: should be distinct from the main source?

presumably if the author wanted both to be treated as one source, they would update the version on DMsGuild

TheGiddyLimit (changes requested):

class/Benjamin Huffman; The Pugilist - 2024.json
```diff
@@ -596,6 +596,12 @@
 					}
 				}
 			],
+			"subclassSpells": [
+				{
+					"className": "Warlock",
+					"classSource": "XPHB"
+				}
+			],
```
nitrodynamite18, line 605: The subclass should only have the Warlock cantrips and 1st-level spells added to the "subclassSpells" section, but I don't know of a way to specify the spell levels, so this is what I'm putting for now.

class/Benjamin Huffman; The Pugilist - 2014.json
```diff
@@ -1,17 +1,18 @@
 {
+	"$schema": "https://raw.githubusercontent.com/TheGiddyLimit/5etools-utils/master/schema/brew-fast/homebrew.json",
 	"_meta": {
 		"sources": [
 			{
 				"json": "SterlingVermin",
```
nitrodynamite18, line 6: This should get changed to something like "BHPugilist" to match the Magus class. Similarly, the Patreon source should have its JSON changed to "BHPugilistPatreon". However, it never ends up working when I try to alter a file's JSON property, so I'll leave it for someone who knows what they're doing.

TheGiddyLimit: changing the `json` part of a source is a lot of faffing about for something which isn't user-visible, so I wouldn't particularly worry about it

consistency is nice, but not breaking things is also nice

TheGiddyLimit (approved):

PR #2748, opened 2026-05-25
img url updates

itsatraaaaap:
Updated img urls to homebrew-img standard.


TheGiddyLimit (changes requested):
you've removed the `"bookData"`? Was this intentional?

PR #2755, opened 2026-06-03
Color of Death

thomasjmarcus-netizen:


TheGiddyLimit (changes requested):
- filename should be e.g. `creature/Author Name; Color of Death.json`
- `siteVersion` should be removed from the JSON

PR #2771, opened 2026-06-12
LaserLlama hotfix

nitrodynamite18:
Fixes a loading error.


revilowaldow:
In #2753 you removed lots of dependencies. Can you put them all back please. Not just those for Wizard.

TheGiddyLimit:
hopefully it's only `PHB` which was actually missing/needed; now added back

PR #2773, opened 2026-06-13
Cantrips and runes

galdevear:


galdevear:
The error is from a file that this commit is not changing. The issue appears to be this commit which is already on main
image

TheGiddyLimit (approved):

PR #2787, opened 2026-07-12
Please can you add this feats to the e5tools?

Dr-BernardRieux:


TheGiddyLimit (changes requested):
I'd love to, but you'll need to first:

- change the filenames to match the repo standard (`Author Name; Homebrew Name.json`)
- ensure all the files are valid JSON (e.g. paste into https://jsonlint.com/ and check)

Dr-BernardRieux:
I did, I sent 4, the 2 first ones didn't had the correct formatting which,
my bad i am really sorry for wasting your time but the second ones where
passed thru a json verifier and had the correct naming scheme, I am sorry
but I am really new to github, heck I dont even study programation I am in
med school there should be 2 pull options from my part with the correct
formating

Dr-BernardRieux:
The thing is I am ao bad at github I cant even tell where to find them

TheGiddyLimit:
I see; so there should be two files only in this PR? Looks like all four are still there, which will be the problem: https://github.com/TheGiddyLimit/homebrew/pull/2787/changes

Unsure how one fixes that from the web UI; I haven't used it much.

If you hit `.` it opens a full editor, though; from there, you could likely find/delete the extra files from the files sidebar, then commit/push from the commit sidebar

Dr-BernardRieux:
there, i deleted both of the incorrectly formatted files, and now both of
the remaining json files are both formatted correctly if i am not incorrect
authors name; of the feat in this case and are valid since i checked them
before uploading them to jsonlit, is there is any errors tell my, it might
be my bad

Dr-BernardRieux:
there i deleted the file that was causing a mistake, i am sory for wasting your time

TheGiddyLimit (approved):

PR #2799, opened 2026-07-29
Create Brennan Lee Mulligan, Dimension 20; Dimension 20.json

nitrodynamite18:
Currently only contains the _Adaine's Furious Fists_ spell, but it can be updated to include all Dimension 20 content.


collection/Brennan Lee Mulligan, Dimension 20; Dimension 20.json
```diff
@@ -0,0 +1,223 @@
...
+				"json": "Dimension20",
+				"version": "",
+				"url": "https://dimension20.fandom.com",
+				"authors": [
+					"Brennan Lee Mulligan"
+				],
+				"convertedBy": [
+					"nitro_dynamite18"
+				],
+				"color": "000000"
```
TheGiddyLimit: might need to see a `colorNight` with that one

TheGiddyLimit (changes requested):

TheGiddyLimit (approved):

TheGiddyLimit:
fwiw if there's no good `version`, putting the current date is customary; lets the next person see when the source material was last retrieved, if nothing else

PR #2816, opened 2026-08-11
Update version and adjust 7 sin feat

DnDMessiah:


DnDMessiah:
If my homebrew contains things other than feats, should I move them to the Collection category?

TheGiddyLimit (approved):

TheGiddyLimit:
> If my homebrew contains things other than feats, should I move them to the Collection category?

Depends; if it's still mostly a "feats" homebrew (e.x., 20 new feats plus 1 supporting spell), then keep it as-is. If the intent is to cover more entity types directly, then moving to collection is appropriate

PR #2831, opened 2026-08-25
A bunch of small fixes

nitrodynamite18:
In the same vein as #2711, removes a bunch of duplicates from some filters and fixes some minor annoyances.


collection/Keith Baker; Exploring Eberron - 2024.json
```diff
@@ -4537,8 +4538,16 @@
 			"classes": {
 				"fromClassList": [
 					{
-						"name": "Artificer",
-						"source": "EFA"
+						"name": "Bard",
```
unknown117343: Please stop adding PHB '14 classes to spells in '24/5.5e sources like this one and Frontiers of Eberron. You do it several times in many sources. Look at PHB'24 spells on 5etools and you will see they also leave out '14 classes on them. I'll leave it to Giddy to decide on the other parts about hardcoding the feats, subclasses, etc.

collection/Keith Baker; Exploring Eberron.json
```diff
@@ -4229,9 +4230,111 @@
 					{
 						"name": "Wizard",
 						"source": "PHB"
+					},
+					{
+						"name": "Sorcerer",
```
unknown117343: You don't need to hardcode this. This JSON is dynamically added to '14 spells already: 

Image

collection/Keith Baker; Exploring Eberron - 2024.json
```diff
@@ -4537,8 +4538,16 @@
...
+						"source": "PHB"
+					},
+					{
+						"name": "Sorcerer",
+						"source": "PHB"
+					},
+					{
+						"name": "Wizard",
+						"source": "PHB"
```
TheGiddyLimit: 5e/'14 class info on spells is auto-generated from the 5.5e/'24 data during runtime; these can be dropped for consistency

TheGiddyLimit (changes requested):

collection/Keith Baker; Exploring Eberron - 2024.json
nitrodynamite18: > 5e/'14 class info on spells is auto-generated from the 5.5e/'24 data during runtime; these can be dropped for consistency

One problem with that is that if someone blocks the 5.5e/'24 classes through the content blocklist filter, the 5e/'14 class isn't added to the list of classes that can learn the spell. For example, blocking the 5.5e/'24 PHB content makes it so the _Magecraft_ spell only gets added to Artificers. Some of my players use blocklists for convenience, so these spells don't get added to all the 5e/'14 stuff that it should and creates confusion.

TheGiddyLimit: sounds like a bug tbh; will look into it

TheGiddyLimit: my mistake; looks like '24 content _does not_ get backwards-migrated to work with '14 content, which is Working As Intended:tm: -- while '24 should be backwards-compatible, '14 should not necessarily be forwards-compatible

so manually adding these is still not desirable, as ultimately, if one is using '14 classes, one should be using '14-compatible homebrew

please ensure:

- '14 class links are not present in '24 homebrew (i.e., the 2024 version of this file)
- '24 class links are not present in '14 homebrew (i.e., the 2014 version of this file)

nitrodynamite18: That doesn't seem like the best solution. Some spells only have a version for '14 or '24, but most spells are compatible with either version of a class by default. Additionally, homebrew subclasses made for SRD '14 classes are automatically applied to SRD '24 classes. I suppose I'm a little confused on 5etools's stance on forwards/backwards compatibility. Is there a written guideline somewhere that I may have missed?

I'll make those changes in the meantime, but I think it's bad practice to say that spells shouldn't be linked to both versions of a class when homebrew subclasses for one version of a class are automatically added to both versions. For testing, you can use my own homebrew file, [Talon's Codex of Secrets](https://github.com/TheGiddyLimit/homebrew/blob/master/collection/nitro_dynamite18%3B%20Talon's%20Codex%20of%20Secrets.json), to see what I mean.

TheGiddyLimit: > Additionally, homebrew subclasses made for SRD '14 classes are automatically applied to SRD '24 classes. 

right, and homebrew spells made for '14 classes automatically receive '24 class links

the supported cases, therefore, are:

- '14 official + '14 homebrew
- '24 official + '24 homebrew
- '24 official + '14 homebrew

the unsupported case is:

- '14 official + '24 homebrew

this is intentional; one should be able to load a '14 brew into `2014.5e.tools` and not have it break because it's missing '24 official content to link to

TheGiddyLimit (approved):

